PR #20, opened 2022-06-14
Support querying in JavaScript via compilation to WASM.

obi1kenobi:
Add the `trustfall_wasm` crate which allows trustfall to be used from web browsers via WASM.


obi1kenobi:
@paavanb if you have a sec, I'd love your thoughts on if the `trustfall_wasm.d.ts` file looks okay.

I'm planning to merge this soon if it looks good, and then:
- hook up GitHub Actions to build the WASM module and make it a (pre-)release artifact on each commit, just like the Python bindings
- convert the example site in the `www` directory to Typescript and actually use the type definitions while I also learn Typescript in the process

PR #29, opened 2022-08-01
Improve browser-based querying UI styles

paavanb:
I was playing around with the OpenAI playground, and when I returned to the browser-based querying demo it looked way too cluttered in comparison. Slimmed it down by dropping the "paper" aesthetic, increasing the contrast on the borders, and decreasing the size of form elements.

image


obi1kenobi:
Looks great! Ship it :)

PR #31, opened 2022-08-05
fix: Clear results on running of query

paavanb:
Since I forgot to clear `results` when the "Run Query" button is pressed, the results get appended to the current results. I also updated to render "Loading..." in the results window while fetching those first results.


eyurtsev (commented):
Looks good

PR #35, opened 2022-08-24
Decouple query playground from HackerNews demo

paavanb:
This PR decouples the HackerNews-specific code from the query playground, which will allow us to create new demos against different data sources while reusing the same underlying components. Specifically, this PR introduces the `TrustfallPlayground` component which is data-agnostic, and the `playgrounds/HackerNews` component which uses that component to render the HN demo.


obi1kenobi (approved):
I don't know enough React to give a useful review, but this sounds great. I trust your judgment, so if you're happy with it, ship it!

PR #41, opened 2022-08-31
Fix playground error bug

paavanb:
Silly mistake: I forgot to clear the error state, so the playground would get stuck after showing an error.


experiments/browser_based_querying/src/rustdoc/Playground.tsx
```diff
@@ -57,9 +57,10 @@ function Playground(props: PlaygroundProps): JSX.Element {
     switch (msg.type) {
       case 'query-ready':
         setResults(msg.results);
+        setError(null);
```
paavanb, line 60: This is the fix, everything else is cosmetic.

PR #48, opened 2022-09-01
Disable Run Query if no query, preload page with first example query

paavanb:
image

image

### DIFF
--- a/experiments/browser_based_querying/src/TrustfallPlayground.tsx
+++ b/experiments/browser_based_querying/src/TrustfallPlayground.tsx
@@ -152,7 +152,7 @@ export default function TrustfallPlayground(props: TrustfallPlaygroundProps): JS
     }
 
     if (noQuery) {
-      return 'Write a query or choose an example one from the right';
+      return 'Write a query or load an example';
     }
 
     return '';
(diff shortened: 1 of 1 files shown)


experiments/browser_based_querying/src/TrustfallPlayground.tsx
```diff
@@ -145,6 +145,19 @@ export default function TrustfallPlayground(props: TrustfallPlaygroundProps): JS
...
   const [selectedTab, setSelectedTab] = useState<ResultsTab>('results');
 
+  const noQuery = encodedQuery === '';
+  const disabledMessage = useMemo(() => {
+    if (disabled) {
+      return disabled;
+    }
+
+    if (noQuery) {
+      return 'Write a query or choose an example one from the right';
```
obi1kenobi, line 155: For some reason, "the right" here doesn't sound good to me, because it's mostly above and not to the right of the text. Any ideas?

Maybe:
- "choose an example query from the dropdown" -- but I wish we didn't name the control element
- "choose an example query from the menu" -- but it's not really a menu

Something like that.

paavanb: "Load an Example Query" is pretty loud, I could mimic that same text to emphasize it. "Write a query or load an example"

paavanb: This will also only appear if they manually clear the query editor, since the page will auto-load the first query.

PR #56, opened 2022-09-05
Resolve a few FE warnings

eyurtsev:
Cleans up a bit of code and resolves two of the warnings seen on the FE

### DIFF
--- a/experiments/browser_based_querying/src/components/DocExplorer.tsx
+++ b/experiments/browser_based_querying/src/components/DocExplorer.tsx
@@ -125,9 +125,7 @@ export function DocExplorer(props: DocExplorerProps) {
           )}
         </Grid>
         <Grid item container direction="column">
-          <Typography>
-            {content}
-          </Typography>
+          {content}
         </Grid>
       </Grid>
     </Grid>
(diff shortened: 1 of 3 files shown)


experiments/browser_based_querying/src/components/DocExplorer.tsx
```diff
@@ -125,9 +125,7 @@ export function DocExplorer(props: DocExplorerProps) {
           )}
         </Grid>
         <Grid item container direction="column">
-          <Typography>
-            {content}
```
paavanb, line 129: This is what fixes the font issue though :(
What error was this causing?

paavanb: Oh is this the `p` problem? You can fix it by doing `<Typography component="div">`, that way it doesn't use a `p` tag.

eyurtsev: This is the `p` warning/error that appears in the log. What's the rationale for including a <Typography> node at this level? `{content}` always comes wrapped in an appropriate node?

paavanb: It turns out it doesn't, sometimes it's unwrapped and results in unstyled text. Putting this at the top level ensures we always use the correct font styles no matter what we render!

paavanb (approved):

experiments/browser_based_querying/src/components/DocExplorer.tsx
eyurtsev: Ah makes sense

PR #57, opened 2022-09-06
Remove HackerNews prefixes from HN demo schema.

obi1kenobi:


obi1kenobi:
@paavanb @eyurtsev this removes the `HackerNews` prefix from all types in the browser demo's HN example, so it might break any outstanding URLs you might have. I think this makes queries look nicer and I'd rather do this break sooner rather than later.

PR #69, opened 2022-09-13
Render error message for Safari

paavanb:
This error only shows for the HackerNews demo, since it relies on passing around `SharedArrayBuffer` instances.


obi1kenobi (approved):
LGTM with one capitalization nit.

PR #75, opened 2022-09-19
Support mobile layout

paavanb:
This PR adds support for a mobile-friendly layout, arranging all editors into a tab-based layout for extra-small and small devices. This was tricky to accomplish, as switching between layouts naively would cause the editors to unmount and lose all their state. Fortunately, the fantastic [react-reverse-portal](https://github.com/httptoolkit/react-reverse-portal) library allows us to render the editors in portals, which are then seamlessly moved around the DOM without forcing re-renders or remounts around layout breakpoints.

## Screenshots
image

image

### DIFF
--- a/experiments/browser_based_querying/package-lock.json
+++ b/experiments/browser_based_querying/package-lock.json
@@ -21,6 +21,7 @@
         "monaco-graphql": "^1.1.2",
         "react": "^18.1.0",
         "react-dom": "^18.1.0",
+        "react-reverse-portal": "^2.1.1",
         "react-router-dom": "^6.3.0",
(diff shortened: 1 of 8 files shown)


obi1kenobi:
I'm pumped, this is awesome! 🙌 🚀

For bonus points, could we reduce the padding left of the line numbers in the query editor so more of the query can be visible? Maybe also reduce the spacing between the right (also the left, but less pressing) edge of the editor and the edge of the screen to a bare minimum?

obi1kenobi:
> Should we hide line numbers entirely on mobile?

If it's easy to do, I say do it. The query staying visible is 1000x more important than the line numbers.

PR #82, opened 2022-09-30
Support deep linking to crates in rust demo

paavanb:
Getting the crate into the URL was the easy part thanks to [use-query-params](https://www.npmjs.com/package/use-query-params), but this introduced a race condition as we would attempt to load the crate before the worker was ready.  I updated the worker to fire a `ready` event so that we only send the load request after the worker was ready.


obi1kenobi:
This is great! 💯 Can we easily associate queries with particular crates as well, so that when the user selects an example query (or on initial load of `/rustdoc`), the correct crate is auto-selected as well?

PR #83, opened 2022-10-02
Implement unlimited-depth directly-optimal recursion.

obi1kenobi:
Implement a new directly-optimal recursion execution algorithm that can support unlimited-depth recursion.

Traces show that the new implementation produces the first result in less than half as many steps on even the shallowest of recursion situations. The more steps previously involved in getting the first result from the recursion, the more significant the improvement will be. 

After this merges, we'll still need to update the front-end to make the `@recurse` depth parameter nullable and plumb that change throughout Trustfall to enable unlimited-depth recursion.

### DIFF
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -83,7 +83,7 @@ where
     }
 }
 
-fn perform_coercion<'query, DataToken>(
+pub(super) fn perform_coercion<'query, DataToken>(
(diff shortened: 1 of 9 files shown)


obi1kenobi:
Unlimited-depth recursion makes adapters even more likely to accidentally trigger the same re-entrancy bug as in #205.

The best way I can see is making a change to the `Adapter` trait to take `&self` instead of `&mut self`, and removing `RefCell` throughout. If users need mutability, they can use their own `RefCell` and will need to take care to avoid re-entrancy issues.

PR #131, opened 2023-02-01
Add a draft interpreter execution hints API.

obi1kenobi:


obi1kenobi:
This is shipping in v0.4.0, so we don't need this prototype anymore.

PR #134, opened 2023-02-09
Ensure errors in filter argument names produce appropriate error messages

ginger51011:
Before, the error simply stated that filter argument names must start with an ASCII letter or underscore character; However what the check did was ensuring the first character was not a number, and then that the whole name only contained ASCII alphanumericals and underscores.

However, this means that the name `"$1example"` would give the error message that ASCII letters and underscores are required at the start, but so would `"$exampleåäö"`, whereas the latter in actuallity is a problem with `åäö` not being ASCII.


trustfall_core/src/resources/test_data/parse_errors/invalid_filter_argument_name.graphql.ron
```diff
@@ -0,0 +1,12 @@
...
+    query: r#"
+{
+    OriginDirectory {
+        name @output @filter(op: "has_substring", value: ["$containsåäö"])
+    }
+}"#,
+    arguments: {
+        "containsåäö": String("src")
+    }
+)
```
obi1kenobi, line 12:
```suggestion
)

```

trustfall_core/src/resources/test_data/parse_errors/invalid_filter_argument_name.parse-error.ron
```diff
@@ -0,0 +1,6 @@
+Err(OtherError(
+    "Filter argument names must only contain ASCII alphanumerics or underscore characters: containsåäö",
+    Pos(
+        line: 4,
+        column: 58,
+)))
```
obi1kenobi, line 6:
```suggestion
)))

```

trustfall_core/src/resources/test_data/parse_errors/invalid_filter_argument_name_start.graphql.ron
```diff
@@ -0,0 +1,12 @@
...
+    query: r#"
+{
+    OriginDirectory {
+        name @output @filter(op: "has_substring", value: ["$1forexample"])
+    }
+}"#,
+    arguments: {
+        "1forexample": String("src")
+    }
+)
```
obi1kenobi, line 12:
```suggestion
)

```

trustfall_core/src/resources/test_data/parse_errors/invalid_filter_argument_name.parse-error.ron
obi1kenobi: Probably possible to configure your editor to add trailing newlines automatically. When they aren't added, some of the unix tools don't work correctly.

trustfall_core/src/resources/test_data/parse_errors/invalid_filter_argument_name_start.parse-error.ron
```diff
@@ -0,0 +1,6 @@
+
+Err(OtherError("Filter argument names must start with an ASCII letter or underscore character: 1forexample",
+Pos(
+  line: 4,
+  column: 58,
+)))
```
obi1kenobi, line 6:
```suggestion
Err(OtherError(
  "Filter argument names must start with an ASCII letter or underscore character: 1forexample",
  Pos(
    line: 4,
    column: 58,
  )
))
```

obi1kenobi (approved):
This is awesome, thank you so much! 🎉

obi1kenobi:
I will fix the clippy errors, they are not caused by this PR.

obi1kenobi:
Your editor appears to not be stripping trailing whitespace on lines, and not adding a final newline at the ends of files, part of which is causing the lint failure. I can take care of this PR while you take a quick look at your editor settings -- these are things it can almost certainly be configured to do automatically for you on file save.

PR #136, opened 2023-02-09
WIP: Add more documentation to `trustfall_core`

ginger51011:
This is a very WIP project related to #135.

Right now there isn't much documentation at all, so I thought I might help get the `rustdoc` at least a bit more easy to navigate. It's a bit write-as-I-go, but I'm open to feedback and suggestions from anyone!

PR #138, opened 2023-02-10
Add some documentation to graphql_query/ in `trustfall_core`

ginger51011:
As discussed #136, this adds some minor documentation to graphql_query. There are still things to do, but these are the things I am somewhat confident writing about.

### DIFF
--- a/trustfall_core/src/graphql_query/directives.rs
+++ b/trustfall_core/src/graphql_query/directives.rs
@@ -254,7 +254,7 @@ impl TryFrom<&Positioned<Directive>> for OutputDirective {
 }
 
 /// A GraphQL `@transform` directive.
-/// 
+///
 /// For example, the following GraphQL and Rust would be equivalent:
 /// ```graphql
 /// @transform(op: "count")
--- a/trustfall_core/src/graphql_query/directives.rs
+++ b/trustfall_core/src/graphql_query/directives.rs
@@ -11,7 +11,7 @@ use crate::ir::{Operation, TransformationKind};
 
 use super::error::ParseError;
 
-/// A value passed as an operator argument in a Trustfall query, for example as in 
+/// A value passed as an operator argument in a Trustfall query, for example as in
 /// the `value` array of the `@filter` directive (see [FilterDirective]).
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum OperatorArgument {
--- a/trustfall_core/src/graphql_query/directives.rs
+++ b/trustfall_core/src/graphql_query/directives.rs
@@ -35,9 +35,9 @@ pub enum OperatorArgument {
 ///
 /// and
 ///
-/// ```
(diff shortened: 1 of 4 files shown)


ginger51011:
I should have ran the doctests...

Marked as `ignore` since I don't think it's possible to access the directive structs from doctests without changing their visibility. This means someone should probably check that the examples are actually correct themselves

trustfall_core/src/graphql_query/directives.rs
```diff
@@ -357,6 +422,7 @@ impl TryFrom<&Positioned<Directive>> for OptionalDirective {
     }
 }
 
+/// A GraphQL `@fold` directive.
```
obi1kenobi, line 425:
```suggestion
/// A Trustfall `@fold` directive.
```

trustfall_core/src/graphql_query/directives.rs
```diff
@@ -377,6 +443,18 @@ impl TryFrom<&Positioned<Directive>> for FoldDirective {
     }
 }
 
+/// A GraphQL `@recurse` directive.
+///
+/// For example, the following GraphQL and Rust would be equivalent:
```
obi1kenobi, line 448:
```suggestion
/// A Trustfall `@recurse` directive.
///
/// For example, the following Trustfall and Rust would be equivalent:
```

trustfall_core/src/graphql_query/error.rs
```diff
@@ -1,3 +1,4 @@
+//! Errors from parsing GraphQL
```
obi1kenobi, line 1:
```suggestion
//! Errors from parsing Trustfall queries
```
Because there are also errors that can come up when parsing Trustfall schemas too.

trustfall_core/src/graphql_query/mod.rs
```diff
@@ -1,3 +1,4 @@
+//! Translating GraphQL queries to Rust types
```
obi1kenobi, line 1: Perhaps a bit more "signposting in global coordinates" here? Maybe something like:
```suggestion
//! Parsing Trustfall queries into Rust types,
//! which are then handed to the frontend for further processing.
```
What do you think?

obi1kenobi (commented):
Looks quite good! Just a few minor polishing suggestions. Thanks for putting this together!

trustfall_core/src/graphql_query/directives.rs
```diff
@@ -1,3 +1,5 @@
+//! Directives in GraphQL can be identified by staring with `@`. While
+//! `trustfall_core` doesn't support all GraphQL directives, some are available.
```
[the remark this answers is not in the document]
ginger51011: Yes, `graphql_query` definitely was a source of confusion for me, since it imples that GraphQL is a core part of Trustfall (it is in `trustfall_core/src` after all!). But thanks for suggestions!

PR #139, opened 2023-02-10
Add minimal documentation to in `ir` in `trustfall_core`

ginger51011:
Again from #136. These are just minimal docs, to help with understanding ` trustfall_core/src/ir/`

### DIFF
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -611,7 +611,7 @@ fn compute_fold<'query, DataToken: Clone + Debug + 'query>(
 ///
 /// A small subtlety is important here: it's possible that the tagged value is *local* to
 /// the scope being filtered. In that case, the context *will not* yet have a token associated
-/// with the vertex ID of the tag's ContextField. However, in such cases, the tagged value
+/// with the [Vid] of the tag's ContextField. However, in such cases, the tagged value
 /// is *never* optional relative to the current scope, so we can safely return `false`.
 #[inline(always)]
 fn is_tag_optional_and_missing<'query, DataToken: Clone + Debug + 'query>(
--- a/trustfall_core/src/ir/mod.rs
+++ b/trustfall_core/src/ir/mod.rs
@@ -307,7 +307,6 @@ impl Argument {
 /// name @filter(op: "=", values: ["$input"])
 /// ```
 /// would produce the `Operation::Equals` variant, for example.
-/// ```
 #[non_exhaustive]
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Operation<LeftT, RightT>
--- a/trustfall_core/src/ir/mod.rs
+++ b/trustfall_core/src/ir/mod.rs
@@ -1,3 +1,4 @@
+//! Trustfall internal representation (IR)
 #![allow(dead_code)]
 
 pub mod indexed;
@@ -26,8 +27,10 @@ lazy_static! {
     pub(crate) static ref TYPENAME_META_FIELD_ARC: Arc<str> = Arc::from(TYPENAME_META_FIELD);
 }
 
+/// Vertex ID
+#[doc(alias("vertex", "node"))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct Vid(pub(crate) NonZeroUsize); // vertex ID
+pub struct Vid(pub(crate) NonZeroUsize);
 
 impl Vid {
     pub fn new(id: NonZeroUsize) -> Vid {
@@ -35,8 +38,10 @@ impl Vid {
     }
 }
 
+/// Edge ID
+#[doc(alias = "edge")]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct Eid(pub(crate) NonZeroUsize); // edge ID
+pub struct Eid(pub(crate) NonZeroUsize);
 
 impl Eid {
     pub fn new(id: NonZeroUsize) -> Eid {
@@ -49,8 +54,12 @@ pub struct EdgeParameters(
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")] pub BTreeMap<Arc<str>, FieldValue>,
 );
 
+/// IR of components of a query, containing information about the vertex ID
+/// of the root of the query, as well as well as maps of all vertices, edges,
+/// folds, and outputs of the query.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct IRQueryComponent {
+    /// The [Vid] of the root, or entry point, of the query.
     pub root: Vid,
 
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
@@ -66,6 +75,7 @@ pub struct IRQueryComponent {
     pub outputs: BTreeMap<Arc<str>, ContextField>,
 }
 
+/// Intermediate representation of a query
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct IRQuery {
     pub root_name: Arc<str>,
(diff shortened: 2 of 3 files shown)


ginger51011:
`rustfmt` fail was on me, my bad

trustfall_core/src/ir/mod.rs
```diff
@@ -49,8 +54,12 @@ pub struct EdgeParameters(
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")] pub BTreeMap<Arc<str>, FieldValue>,
 );
 
+/// IR of components of a query, containing information about the vertex ID
+/// of the root of the query, as well as well as maps of all vertices, edges,
+/// folds, and outputs of the query.
```
obi1kenobi, line 59:
```suggestion
/// A complete component of a query; may itself contain one or more components.
///
/// Contains information about the Vid where the component is rooted,
/// as well as well as maps of all vertices, edges, folds, and outputs from this component.
```
What do you think?
- Is the "one-liner explanation at the top" a good pattern, both for docs.rs and for IDEs?
- I added a bit more explanation to differentiate how query component and query are different.

trustfall_core/src/ir/mod.rs
```diff
@@ -49,8 +54,12 @@ pub struct EdgeParameters(
...
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct IRQueryComponent {
+    /// The [Vid] of the root, or entry point, of the query.
```
obi1kenobi, line 62:
```suggestion
    /// The [Vid] of the root, or entry point, of the component.
```
Because it can be the entry point of a subcomponent, not the top-level one. If you take a peek at `IRFold` (the one that's part of the `folds` field), you'll find another `IRQueryComponent` in there.

trustfall_core/src/ir/mod.rs
```diff
@@ -1,3 +1,4 @@
+//! Trustfall internal representation (IR)
```
obi1kenobi, line 1:
```suggestion
//! Trustfall intermediate representation (IR)
```
Though it can stand for both meanings for sure. I prefer "intermediate" slightly because it doesn't give the accidental impression that the IR is meant to only be readable inside Trustfall itself, which is counter to the vision of having a universal data layer: "LLVM for data sources."

trustfall_core/src/ir/mod.rs
```diff
@@ -94,6 +104,9 @@ pub struct IREdge {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub parameters: Option<Arc<EdgeParameters>>,
 
+    /// Indicating if this edge is optional.
+    ///
+    /// This would correspond to `@optional` in GraphQL.
```
obi1kenobi, line 109:
```suggestion
    /// Corresponds to the `@optional` directive.
```
I mentioned this in #142 -- let's try to uphold the distinction between Trustfall and GraphQL as much as possible, and further distinguish over time.

trustfall_core/src/ir/mod.rs
```diff
@@ -282,6 +299,15 @@ impl Argument {
...
 }
 
+/// Operations that can be made in the graph.
+///
+/// In GraphQL, this can correspond to the `op` argument in `@filter`,
+/// for example in the following:
+/// ```graphql
+/// query Student {
+///     name @filter(op: "has_substring", values: ["John"])
+/// }
```
obi1kenobi, line 309:
```suggestion
/// In a Trustfall query, the `@filter` directive produces `Operation` values:
/// ```graphql
/// name @filter(op: "=", values: ["$input"])
/// ```
/// would produce the `Operation::Equals` variant, for example.
```
Because the other query wasn't actually valid:
- It has a property field on the root node, which in Trustfall should produce a schema error when attempting to parse it. (Not sure this is tested! If not, probably a great entry point for contributing some code, if you're interested, since you almost certainly know more GraphQL than I do!)
- `@filter` can't take literal values at the moment, only `"$foo"` runtime arguments or `"%foo"` tagged arguments (see the `@tag` directive, it "captures" the tagged value and allows subsequent filters to refer back to it).

I felt that a fully-valid query would be too complex (at least two more lines), so I went with an excerpt. What do you think?

trustfall_core/src/ir/mod.rs
```diff
@@ -365,6 +391,10 @@ where
         }
     }
 
+    /// The operation name as a `str`
+    ///
+    /// Note that these are the same as would be given to a GraphQL `op`
+    /// argumetn.
```
obi1kenobi, line 397: Let's just merge these? The return type is in the function signature so I don't think we need to restate it.
```suggestion
    /// The operation name, as it would have appeared in the `@filter` directive `op` argument.
```

trustfall_core/src/ir/value.rs
```diff
@@ -1,7 +1,11 @@
+/// IR of the values of GraphQL fields.
 use async_graphql_value::{ConstValue, Number, Value};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
+/// Values of fields in GraphQL types.
```
obi1kenobi, line 6:
```suggestion
/// Values of fields in Trustfall.
```

trustfall_core/src/ir/value.rs
```diff
@@ -1,7 +1,11 @@
+/// IR of the values of GraphQL fields.
```
obi1kenobi, line 1:
```suggestion
/// IR of the values of Trustfall fields.
```

trustfall_core/src/ir/value.rs
```diff
@@ -1,7 +1,11 @@
...
 use async_graphql_value::{ConstValue, Number, Value};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 
+/// Values of fields in GraphQL types.
+///
+/// For version that is serialized as an untagged enum, see [TransparentValue].
```
obi1kenobi, line 8: awesome!

obi1kenobi (commented):
Overall, looks really good -- thanks for putting it together! Just a few polishing suggestions here and there. Let me know what you think?

trustfall_core/src/ir/mod.rs
```diff
@@ -49,8 +54,12 @@ pub struct EdgeParameters(
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")] pub BTreeMap<Arc<str>, FieldValue>,
 );
 
+/// IR of components of a query, containing information about the vertex ID
```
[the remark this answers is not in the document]
ginger51011: I think it might be worth linking it every  time (so it's always easy to find the definition), but yes it should be the same. I'll replace them!

trustfall_core/src/ir/mod.rs
ginger51011: Looking at some `std` docs, it seems to be pretty standard? I do it because when you then look in the lists of for example structs (like [here](https://doc.rust-lang.org/std/sync/index.html#structs)), you get the same one-liner as a brief explanation, so you can see _Ooh that's what I want!_.

I suggest we follow this pattern (which will mean no periods at the end of the one-liner. What do you think?

Also thanks for the explanation!

trustfall_core/src/ir/mod.rs
ginger51011: I agree, this is better. I will look into the property field on the root node!

ginger51011:
Thank you for taking the time to review!

trustfall_core/src/interpreter/execution.rs
```diff
@@ -611,7 +611,7 @@ fn compute_fold<'query, DataToken: Clone + Debug + 'query>(
 ///
 /// A small subtlety is important here: it's possible that the tagged value is *local* to
 /// the scope being filtered. In that case, the context *will not* yet have a token associated
-/// with the vertex ID of the tag's ContextField. However, in such cases, the tagged value
+/// with the [Vid] of the tag's ContextField. However, in such cases, the tagged value
```
obi1kenobi, line 614: Doesn't have to be in this PR, just wondering: can we link `ContextField` here too?

trustfall_core/src/ir/mod.rs
```diff
@@ -282,6 +300,14 @@ impl Argument {
...
 }
 
+/// Operations that can be made in the graph.
+///
+/// In a Trustfall query, the `@filter` directive produces `Operation` values:
+/// ```graphql
+/// name @filter(op: "=", values: ["$input"])
+/// ```
+/// would produce the `Operation::Equals` variant, for example.
+/// ```
```
obi1kenobi, line 310:
```suggestion
```

obi1kenobi (approved):

obi1kenobi:
Thanks for all the awesome contributions, they are making a huge difference 🚀

PR #140, opened 2023-02-10
Add minimal frontend docs

ginger51011:
Final one for now (see #136) :)

### DIFF
--- a/trustfall_core/src/frontend/mod.rs
+++ b/trustfall_core/src/frontend/mod.rs
@@ -1,4 +1,4 @@
-//! Frontend for Trustfall, containing parsers for queries using schemas
+//! Frontend for Trustfall: takes a parsed query, validates it, and turns it into IR.
 #![allow(dead_code, unused_variables, unused_mut)]
 use std::{
     collections::BTreeMap, convert::TryFrom, iter::successors, num::NonZeroUsize, sync::Arc,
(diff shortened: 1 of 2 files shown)


trustfall_core/src/frontend/mod.rs
```diff
@@ -41,6 +42,9 @@ mod tags;
 mod util;
 mod validation;
 
+/// Parses a query string to the Trustfall IR using a provided
+/// [Schema](crate::schema::Schema). May fail if [parse_to_ir](parse_to_ir)
+/// fails for the provided schema and query.
```
obi1kenobi, line 47: I'm not 100% sure on what the recommended style for Rust doc comments is. My impression is that imperative mood ("Parse a query string...") is somewhat common, and I have a slight preference for it since it's what I'm used to.

But if the Rust style docs say otherwise, I'm happy to adhere to the community norms. Same if there's another strong reason to not use imperative.

What do you think?

ginger51011: I had to look myself, and looking at the docs for [tokio](https://docs.rs/tokio/1.25.0/tokio/#macros) and [std](https://doc.rust-lang.org/stable/core/mem/index.html#functions) it seems like my suggested form is somewhat standard. I won't say I have a strong preference either way though :shrug:

obi1kenobi: Neat, let's stick with the standard! Thanks, I definitely would have done it backwards then.

Issue #141, opened 2023-02-10
Moving or renaming testcases in `trustfall_core/src/resources`


ginger51011:
I noticed when I was writing my first PR that it is not completely obvious what or where the tests in `trustfall_core` are. Right now they can be found in `trustfall_core/src/resources`, but without any README this isn't at all obvious (after all, `resources` could just be a list of constants). Is there a motivation that I am missing (and if so, could it be clarified).

For me its a bit odd that test data is provided in a `src/` directory; It is not used in the final crate (as far as I know), with the exception of perhaps the schemas.

Would it make sense to move parameterized testcases to for example `trustfall_core/testcases`? Or would a README with directory structure explained in `trustfall_core` be enough? 

This is in part related to #135

obi1kenobi:
Great point. No good-enough reason.

Once #143 merges, it's safe to move them. How about `trustfall_core/test_data`?

PR #144, opened 2023-02-11
Move core testcases

ginger51011:
See #141.

Some motivation: Now `test_data` is directly at crate root, where one could expect to find it. Not naming it `tests` here prevents `cargo` expecting to find integration tests there.

The nested `test_data` (with the input-expected output pairs) have been renamed to `test_data/tests`, because well, they are tests (this could be up for debate). I also moved the test README to `trustfall_core/test_data/README.md` because the schemas themselves are part of the test data (or so I would say?), and I think this makes it less hidden. I also updated the README with links to directories, so it points where to put new tests. Also some rewrapping because why not.

Tests pass locally after I changed the paths.

Sidenote:
I have noticed in `main.rs` and at other points is the handling of paths as strings. I'm not sure, but could this be a problem for Windows users? `std::path::Path` might be a good alternative (I have tried to get it to work, but it got a bit messy so maybe not worth it).


obi1kenobi (approved):

obi1kenobi:
> I have noticed in main.rs and at other points is the handling of paths as strings. I'm not sure, but could this be a problem for Windows users? `std::path::Path` might be a good alternative (I have tried to get it to work, but it got a bit messy so maybe not worth it).

Yes, right now the repo assumes an Unix-y OS with its choice of scripts and path handling. Developing on Windows would make everything pretty horribly broken ... which is ironic because I do all my dev work on a Windows machine, under WSL2 :)

I think *right now* it isn't worth it, in that it would be a relatively significant lift for an unclear payoff, and I'd rather focus on docs and the other reorganization and other ergonomics improvements first. But I'll open an issue so we don't forget about it, and we can circle back in the future.

PR #154, opened 2023-02-12
Apply all the major breaking changes to Adapter and surrounding code

obi1kenobi:
- Rename all the `Adapter` methods, lifetime, and associated type to match the new names in `BasicAdapter`.
- Move the `ContextIterator, ContextOutcomeIterator, VertexIterator` to `trustfall_core::interpreter`.
- Use the above type aliases wherever possible.

### DIFF
--- a/demo-feeds/src/adapter.rs
+++ b/demo-feeds/src/adapter.rs
@@ -2,8 +2,9 @@ use feed_rs::model::{Content, Entry, Feed, FeedType, Image, Link, Text};
 use trustfall_core::{
     field_property,
     interpreter::{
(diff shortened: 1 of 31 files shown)


obi1kenobi:
Two more sets of breaking changes I still want to do before we release `0.2.0` whose `Adapter` trait should be reasonably stable:
- Eliminate all the near-useless "hints" arguments in `Adapter` and replace them with a pretty bare `QueryInfo` type, which will be the foundation of the upcoming adapter-optimization functionality mentioned in my "2000x speedup" blog post.
- Clean up the mess that is the `EdgeParameters` type to make it more ergonomic.

PR #160, opened 2023-02-14
Makes links in README relative file links, fix some of them

ginger51011:
I think it makes more sense to link to local files than to link to GitHub. Also, some of them were outdated so I fixed them.

GitHub has support for this, see https://github.blog/2013-01-31-relative-links-in-markup-files/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ etc. For a 10min video introduction to the project, see
 the ["How to Query (Almost) Everything" talk](https://www.hytradboi.com/2022/how-to-query-almost-everything)
 from the [HYTRADBOI 2022](https://www.hytradboi.com/) conference.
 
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Thanks for the fixes!

My only concern is whether the relative links will also work correctly when other services (e.g. crates.io here: https://crates.io/crates/trustfall_core ) parse the README. Is that something you've already had a chance to look into?

ginger51011:
I had actually not thought of that, but it seems like something crates.io resolves according to [this](https://github.com/rust-lang/crates.io/pull/3973)?

obi1kenobi (approved):

obi1kenobi:
Amazing, thanks for looking into it!

PR #167, opened 2023-02-14
Document `DataContext` and its `active_vertex()` method.

obi1kenobi:


obi1kenobi:
@ginger51011 I'm planning to merge this quickly since "something is better than nothing" and this is very user-facing code. As always, if you get a chance to take a look and have any suggestions for improvement, please flag them and/or directly open a PR. No pressure to look at all!

ginger51011:
Looking through your changes, it all looks great! I like the move towards using the same language everywhere with `Vertex`, makes it less confusing at first glance.

obi1kenobi:
Agreed, and now is definitely a good time to do it. I think the `DataContext` internals are some of the last places that still need to change to the new nomenclature, and I'm about to do that next.

PR #171, opened 2023-02-15
Extract a more ergonomic `trustfall` crate, and publish v0.2.0 of major crates.

obi1kenobi:


obi1kenobi:
Still a draft — planning to actually try using this in one of the example implementations before merging.

But @ginger51011 would still love your thoughts if you have a sec, for the general idea / layout / naming of things.

ginger51011:
This is a welcome change!

I think it is a good idea to move the demos with this into a `trustfall/examples/` directory, and link add some information about it in the README.

Also, is it a good idea to extract `helper.rs` and the macros to just the `trustfall` crate? It might make it more fractured, but at the same time is it really considered to be "core" functionality? I'm not convinced either way to be honest.

obi1kenobi:
> I think it is a good idea to move the demos with this into a `trustfall/examples/` directory, and link add some information about it in the README.

Agreed. I might do that in a separate PR, just to keep PR sizes down.

> Also, is it a good idea to extract `helper.rs` and the macros to just the `trustfall` crate? It might make it more fractured, but at the same time is it really considered to be "core" functionality? I'm not convinced either way to be honest.

I'm open to moving them out of "core" for sure. At the same time, I think it would be bad if we force users to have to install `trustfall` *and* `trustfall_core` to get the "full" functionality, because that might end up pulling in two different major versions of `trustfall_core` (one directly, and one as a dependency of `trustfall`) and then cause much frustration.

I also think it's a good idea to re-export the macros in `trustfall` regardless of where they are defined. If you think it's worth defining them outside of "core" then perhaps it's time to make that `trustfall_macros` crate? :)

PR #173, opened 2023-02-16
Add derive macro crate that generates `typename()` and `as_` conversions.

obi1kenobi:
Extension to the idea in #163.

```rust
use trustfall_derive::TrustfallEnumVertex;

#[derive(Debug, Clone, TrustfallEnumVertex)]
enum Vertex {
    User(String),

    #[trustfall(skip_conversion)]
    Message { author: String, content: String },
}
```

The above enum effectively gets the following implementations:
```rust
impl Typename for Vertex {
    fn typename(&self) -> &'static str {
        match self {
            Self::User { .. } => "User",
            Self::Message { .. } => "Message",
        }
    }
}

impl Vertex {
    fn as_user(&self) -> Option<&String> {
        match self {
            Self::User(x) => Some(x),
            _ => None,
        }
    }
}
```
The `as_message()` conversion method was not generated since it was explicitly disabled with `#[trustfall(skip_conversion)]`.


obi1kenobi:
Currently works only for tuple and plain variants. Open work items:
- [ ] UI tests for the error message when the derive macro is applied to a non-enum
- [ ] Should we generate conversion functions for struct variants? Either need UI tests for the error, or tests to make sure it works.

trustfall_proc_macros/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+fn impl_variants_as_vertex_types(ast: &syn::DeriveInput) -> proc_macro::TokenStream {
+    let name = &ast.ident;
+    let generics = &ast.generics;
+    let where_clause = &generics.where_clause;
+
+    let variants = match &ast.data {
+        syn::Data::Enum(d) => &d.variants,
+        _ => panic!("Only enums can become named types!"),
```
ginger51011, line 16: Named types or TrustfallEnumVertex or VariantsAsVertexTypes?

trustfall_proc_macros/src/lib.rs
```diff
@@ -0,0 +1,135 @@
+use quote::quote;
+
+#[proc_macro_derive(VariantsAsVertexTypes)]
```
ginger51011, line 3: Suggestions for trait name: `TrustfallEnumVertex`? Describes what it is, `Trustfall` makes sure of name conflicts, `EnumVertex` makes it clear that yes it is a special type of `Vertex` made of an enum?

ginger51011, line 3: It would be very nice with docs for _Before_ and _After_ this macro, so it's clear what you're actually adding. Reading the code will be of limited help if you don't know what you're looking for.

(Perhaps `cargo-extend` could make this easy)

ginger51011 (commented):
Very nice! Great job

ginger51011:
As for structs/functions: I do see the point, but if so I think it should be another PR. Right now all the demos are made this way, so perhaps it is best to start and see if the need comes up?

trustfall_proc_macros/src/lib.rs
obi1kenobi: Love it!

obi1kenobi:
> As for structs/functions: I do see the point, but if so I think it should be another PR. Right now all the demos are made this way, so perhaps it is best to start and see if the need comes up?

Ah sorry, I think what I wrote was confusing: I meant struct variants of enums, like this:
```rust
enum Vertex {
    StructLike {
        a: usize,
        b: &'static str,
    }
}
```

In this case, `typename()` is easy but the return type of `fn as_struct_like()` is tricky — should it be `Option<(&usize, &&'static str)>` with a tuple of the fields in the same order, or something else?

obi1kenobi:
I realized that sometimes users may want a different conversion method, so now the `#[trustfall(skip_conversion)]` attribute on a variant disables generating a conversion method for it — so that we don't hog the `as_<variant>()` method name and the user can implement it themselves.

Added docs and updated the PR headline to match.

obi1kenobi:
Done tweaking, planning to merge if all looks good to you @ginger51011!

ginger51011:
Oh I really like this! Especially making it skippable, that prevents possible headaches.

I can look closer at the actual implementation later, but from a design perspective this is great!

trustfall_derive/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        quote!((#punct))
+    } else {
+        panic!(
+            "list of fields had {} field(s), which is not more than one field",
+            fields.len()
+        );
+    }
+}
+
+fn to_lower_snake_case(value: &str) -> String {
```
ginger51011, line 334: Perhaps some tests will help verify this. It looks right to me, but it would be nice with some tests for sketchy names, like starting with uppercase, double uppercase, etc.

I saw that there are integration tests for it, but I went looking for them in this file

ginger51011 (commented):
Wow, this is great! I had one small comment and one thought.

If we want to add a trait bound later for `Typename` for `BasicAdapter, we might want to separate this into two macros. From a user point of view, it would be very easy to do two derives, but very annoying to ignore everything if you only want `Typename` functionality. 

Also, when I see a derive, I usually think "Ah I add this trait!". Here we add a trait and also some methods. What do you think?

trustfall_derive/src/lib.rs
obi1kenobi: It's easy enough to add them here too, I don't mind 👍

obi1kenobi:
> If we want to add a trait bound later for `Typename` for `BasicAdapter`, we might want to separate this into two macros. From a user point of view, it would be very easy to do two derives, but very annoying to ignore everything if you only want `Typename` functionality.
> 
> Also, when I see a derive, I usually think "Ah I add this trait!". Here we add a trait and also some methods. What do you think?

Adding a `Typename` bound for `BasicAdapter` is an excellent idea! I'll do that right now.

I'm conflicted on the two derives. I feel like two derives might be a bit too intimidating for new users ("double the new things to learn"), but it definitely does disadvantage slightly more advanced users as you point out.

Could we just add a new `Typename` derive macro that just implements `Typename` for those cases, while keeping the `TrustfallEnumVertex` macro as the "one stop shop" for the full functionality?

ginger51011:
Yes that would give more control! If we just add docstrings with links between the macros, confusion would be kept at a minimum. As you say most users would probably just use `TrustfallEnumVertex`

ginger51011 (approved):
LGTM!

PR #175, opened 2023-02-19
Require `BasicAdapter::Vertex` type to implement `Typename`.

obi1kenobi:
`BasicAdapter` now requires its `Vertex` type to implement `Typename`. In exchange, it automatically handles the type name property while still allowing implementers to override that logic. Default implementations of trait methods for the win!

### DIFF
--- a/demo-feeds/src/adapter.rs
+++ b/demo-feeds/src/adapter.rs
@@ -4,7 +4,7 @@ use trustfall_core::{
     interpreter::{
         basic_adapter::BasicAdapter,
         helpers::{resolve_neighbors_with as neighbors, resolve_property_with as property},
-        ContextIterator, ContextOutcomeIterator, VertexIterator,
+        ContextIterator, ContextOutcomeIterator, Typename, VertexIterator,
     },
     ir::{EdgeParameters, FieldValue},
 };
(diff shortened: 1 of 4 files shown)


demo-feeds/src/adapter.rs
```diff
@@ -50,6 +50,19 @@ impl<'a> Vertex<'a> {
...
+        match self {
+            Vertex::Feed(..) => "Feed",
+            Vertex::FeedText(..) => "FeedText",
+            Vertex::ChannelImage(..) => "ChannelImage",
+            Vertex::FeedLink(..) => "FeedLink",
+            Vertex::FeedEntry(..) => "FeedEntry",
+            Vertex::FeedContent(..) => "FeedContent",
+        }
+    }
+}
```
obi1kenobi, line 64: As soon as we've stabilized the `trustfall` crate's structure, I'm going to move the examples to an `examples/` directory and then make them take advantage of the new derive macros.

Changes like this one in the demo projects are just to make everything compile for the time being.

trustfall_core/src/interpreter/basic_adapter.rs
```diff
@@ -142,6 +149,48 @@ pub trait BasicAdapter<'vertex> {
...
+    ///
+    /// [`DataContext`]: super::DataContext
+    /// [`Schema`]: crate::schema::Schema
+    fn resolve_typename(
+        &mut self,
+        contexts: ContextIterator<'vertex, Self::Vertex>,
+        _type_name: &str,
+    ) -> ContextOutcomeIterator<'vertex, Self::Vertex, FieldValue> {
+        resolve_property_with(contexts, |vertex| vertex.typename().into())
+    }
```
obi1kenobi, line 193: @ginger51011 curious what you think about this design choice: `BasicAdapter` handles typenames automatically but allows that logic to be overridden as well.

obi1kenobi (commented):
I think this is probably one of the last breaking changes before releasing v0.2.0 of `trustfall_core` and the first version of `trustfall`.

trustfall_core/src/interpreter/basic_adapter.rs
ginger51011: I think it's very reasonable, but what is the use of the `_type_name` parameter right now? A quick glance at the docs doesn't explain it and it is a bit confusing to have `resolve_typename` require a typename

obi1kenobi: Good point, as always! I expanded that bit of the docs with a small example. Let me know what you think!

ginger51011: Yes, that makes more sense

ginger51011 (approved):
LGTM :)

trustfall_core/src/interpreter/basic_adapter.rs
obi1kenobi: Awesome! Thanks for checking it out, much appreciated!

PR #201, opened 2023-03-01
Add `From<&Vec<T>>` and `From<&&[T]>` for `FieldValue`.

obi1kenobi:
Type inference doesn't always realize that it can use the `From<&[T]>`
implementation for `&Vec`, so let's help it out.


obi1kenobi:
Superseded by #204.

PR #204, opened 2023-03-01
Tidy From<_> for FieldValue impls

sharnoff:
Mainly removes the `From<&_>` implementations and forces the caller to use `Clone`, etc.

~~Halfway through, I added an impl for `Vec<T>`, so there might be a couple places where there's an unnecessary `vec.into_iter().collect()`.~~ — Fixed

Should hopefully help resolve the problem behind #201 :)

### DIFF
--- a/demo-hytradboi/src/adapter.rs
+++ b/demo-hytradboi/src/adapter.rs
@@ -264,7 +264,7 @@ impl Adapter<'static> for DemoAdapter {
 
             // properties on Webpage
             ("Webpage" | "Repository" | "GitHubRepository", "url") => {
-                impl_property!(contexts, as_webpage, url)
+                impl_property!(contexts, as_webpage, url, { url })
             }
 
             // properties on GitHubRepository
--- a/demo-hytradboi/src/adapter.rs
+++ b/demo-hytradboi/src/adapter.rs
@@ -264,7 +264,7 @@ impl Adapter<'static> for DemoAdapter {
 
             // properties on Webpage
             ("Webpage" | "Repository" | "GitHubRepository", "url") => {
-                impl_property!(contexts, as_webpage, url, { url })
+                impl_property!(contexts, as_webpage, url)
             }
 
             // properties on GitHubRepository
--- a/demo-hytradboi/src/adapter.rs
+++ b/demo-hytradboi/src/adapter.rs
@@ -126,24 +126,20 @@ impl DemoAdapter {
 macro_rules! impl_item_property {
     ($contexts:ident, $attr:ident) => {
         Box::new($contexts.map(|ctx| {
-            let vertex = ctx.active_vertex();
-            let value = match vertex {
-                None => FieldValue::Null,
-                Some(t) => {
+            let value = ctx
+                .active_vertex()
+                .map(|t| {
                     if let Some(s) = t.as_story() {
-                        (&s.$attr).into()
+                        s.$attr.clone()
                     } else if let Some(j) = t.as_job() {
-                        (&j.$attr).into()
+                        j.$attr.clone()
                     } else if let Some(c) = t.as_comment() {
-                        (&c.$attr).into()
+                        c.$attr.clone()
                     } else {
                         unreachable!()
                     }
-                }
-
-                #[allow(unreachable_patterns)]
-                _ => unreachable!(),
-            };
+                })
+                .into();
 
             (ctx, value)
         }))
@@ -156,13 +152,7 @@ macro_rules! impl_property {
             let vertex = ctx
                 .active_vertex()
                 .map(|vertex| vertex.$conversion().unwrap());
-            let value = match vertex {
-                None => FieldValue::Null,
-                Some(t) => (&t.$attr).clone().into(),
-
-                #[allow(unreachable_patterns)]
-                _ => unreachable!(),
-            };
+            let value = vertex.map(|t| t.$attr.clone()).into();
 
             (ctx, value)
         }))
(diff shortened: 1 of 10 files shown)


obi1kenobi:
Does that mean the `field_property!()` and `accessor_property!()` macros should be tweaked like so? (I'll add test cases for this in a later PR, this is just from testing with some toy adapters I have locally.)

Specifically, does `.clone()` when accessing a field but `.to_owned()` when using the output of an accessor function do the right thing? I'm hoping that `.to_owned()` will be a no-op if the accessor returned an owned value already.

```diff
diff --git a/trustfall_core/src/interpreter/helpers.rs b/trustfall_core/src/interpreter/helpers.rs
index e20bc25..b2254b2 100644
--- a/trustfall_core/src/interpreter/helpers.rs
+++ b/trustfall_core/src/interpreter/helpers.rs
@@ -184,21 +184,21 @@ pub fn resolve_coercion_with<'vertex, Vertex: Debug + Clone + 'vertex>(
 macro_rules! field_property {
     // If the data is a field directly on the vertex type.
     ($field:ident) => {
-        |vertex| -> FieldValue { (&vertex.$field).into() }
+        |vertex| -> FieldValue { vertex.$field.clone().into() }
     };
     // If we need to call a fallible conversion method
     // (such as `fn as_foo() -> Option<&Foo>`) before getting the field.
     ($conversion:ident, $field:ident) => {
         |vertex| -> FieldValue {
             let vertex = vertex.$conversion().expect("conversion failed");
-            (&vertex.$field).into()
+            vertex.$field.clone().into()
         }
     };
     // Supply a block to post-process the field's value.
     // Use the field's name inside the block.
     ($conversion:ident, $field:ident, $b:block) => {
         |vertex| -> FieldValue {
-            let $field = &(vertex.$conversion().expect("conversion failed").$field);
+            let $field = &vertex.$conversion().expect("conversion failed").$field;
             $b
         }
     };
@@ -263,14 +263,14 @@ macro_rules! field_property {
 macro_rules! accessor_property {
     // If the data is available as an accessor method on the vertex type.
     ($accessor:ident) => {
-        |vertex| -> FieldValue { (&vertex.$accessor()).into() }
+        |vertex| -> FieldValue { vertex.$accessor().to_owned().into() }
     };
     // If we need to call a fallible conversion method
     // (such as `fn as_foo() -> Option<&Foo>`) before using the accessor.
     ($conversion:ident, $accessor:ident) => {
         |vertex| -> FieldValue {
             let vertex = vertex.$conversion().expect("conversion failed");
-            (&vertex.$accessor()).into()
+            vertex.$accessor().to_owned().into()
         }
     };
```

sharnoff:
Ah, missed that I needed to run clippy with `--all-targets`. I do believe changing the macros in the ways you highlighted is correct — although needs tests, etc :P

`to_owned` takes `&self`, so it unfortunately can't do that "no-op if already owned" thing, so this might introduce unnecessary cloning? I think I've lost the context though - not entirely sure what you're referring to.

(FWIW, if you want to do this in a generic way, `Cow::from(...).into_owned()` might work — for example, `Cow` has implementations of `From<&str>` and `From<String>`, although making that work in macros might be tricky.)

obi1kenobi:
> `to_owned` takes `&self`, so it unfortunately can't do that "no-op if already owned" thing, so this might introduce unnecessary cloning? I think I've lost the context though - not entirely sure what you're referring to.

No worries, I think it isn't a huge deal even if clones unnecessarily.

Here's an example where that's the case (unless the optimizer does some magic): imagine a Trustfall property is defined as using the value of an accessor method `foo(&self) -> String`. The `accessor_property!()` macro will then call `vertex.foo().clone().into()` to make the `FieldValue` there, even though `vertex.foo().into()` would have been sufficient.

The `Cow` idea is a good suggestion, but I think it could be tricky with a `macro_rules` macro. The Trustfall library has to work with however the underlying data source represents the data, so if the underlying data source has a field or accessor that produces `String` (or `&str`) but not a `Cow`, then it'll be difficult for a `macro_rules` macro to turn that into a `Cow`.

Thanks for the help on this! I'll merge it and release soon. I think it's a breaking change so it'll be v0.3.0.

demo-hytradboi/src/adapter.rs
```diff
@@ -285,14 +264,14 @@ impl Adapter<'static> for DemoAdapter {
 
             // properties on Webpage
             ("Webpage" | "Repository" | "GitHubRepository", "url") => {
-                impl_property!(contexts, as_webpage, url, { url.into() })
+                impl_property!(contexts, as_webpage, url, { url })
```
obi1kenobi, line 267:
```suggestion
                impl_property!(contexts, as_webpage, url)
```

ginger51011:
A bit late to the party, but this change created regressions on my part, as `From<&String>` no longer exists. Is this desired @obi1kenobi? It just means a call to `as_str` on my part, but that is not really ergonomic if you ask me.

(It also broke a `From<DateTime<Utc>>`, but I don't think that is as bad, it was a bit obfuscated that I wanted to map over it and get a timestamp)

obi1kenobi:
Hmm, interesting. Perhaps the real answer was "fix my macros" without removing _all_ the `From` impls we removed, specifically not removing the ones for `&T` where `T` is not `Copy`? @sharnoff, would love your thoughts if you have a sec!

I think the above avoids the original problem of needing `From<&&str>` and `From<&i64>` and such, while not requiring extra steps for `&String`.

obi1kenobi:
I think `From<DateTime<Utc>>` was never implemented 😅 I don't think I've tested any of the code paths involving `DateTime<Utc>` so it's possible they might not work properly or might panic unnecessarily.

I'm adding a `From<&String>` implementation again and I'm going to release it shortly.

PR #207, opened 2023-03-03
Make `field_property` macro accept literals for indexing

ginger51011:
Having a `Vertex` variant that contains a tuple, it would be nice to just be able to pass an index (although this would work for maps as well).

I understand that this _kind of_ breaks the naming of `field_property`, but in contrast with `accessor_property` the behaviour here seems to be very close, and making another macro would just be odd to me. Any ideas welcome, I guess it could be renamed but that's perhaps annoying.

TODO: Add some doctests

Edit: Yeah this was more complicated than I thought. I will test a bit to see if I can find a nice way to do this, or if index_property is the right way to go

### DIFF
--- a/trustfall_core/src/interpreter/helpers.rs
+++ b/trustfall_core/src/interpreter/helpers.rs
@@ -76,7 +76,7 @@ pub fn resolve_coercion_with<'vertex, Vertex: Debug + Clone + 'vertex>(
 /// Generally used with [`resolve_property_with`].
 ///
 /// Retrieves a [`FieldValue`] from a vertex by converting it to the proper type,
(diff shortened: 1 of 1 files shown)


ginger51011:
Hmm, no, I don't like this. Nvm.

obi1kenobi:
For what it's worth, I've tried using tuples as vertex types multiple times and always regretted it sooner or later. There's always the need for "just one more field" and field indexes aren't nearly as effective at being self-documenting as named fields.

These days I always use plain structs inside the variant. If your experience ends up being the same, then perhaps we should write this into the docs somewhere.

PR #229, opened 2023-03-31
Allow using `*_property` helper macros without importing `FieldValue`

### LINKED ISSUE #203: `accessor_property` and `field_property` macros should use absolute paths
They currently seem to assume that types like `FieldValue` are directly imported, and can cause errors like:
```
error[E0412]: cannot find type `FieldValue` in this scope
  --> .../src/adapter.rs:43:59
   |
43 |                 "name" => resolve_property_with(contexts, accessor_property!(as_inner, name)),
   |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
   |
   = note: this error originates in the macro `accessor_property` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this enum
   |
1  | use trustfall::FieldValue;
   |

For more information about this error, try `rustc --explain E0412`.
error: could not compile `...` due to previous error
```

ginger51011:
Fixes #203 

Before, the macro would not compile without `use ...::FieldValue` in the same scope. This uses the special `$crate` metavariable to fix this.

Haven't tested this properly, do we have anywhere nice to put the tests?

### DIFF
--- a/trustfall/tests/helper_macros.rs
+++ b/trustfall/tests/helper_macros.rs
@@ -1,0 +1,130 @@
+// Do NOT import FieldValue, we want to check that accessor macros does what
+// we want
+
+struct Adapter;
+
+#[derive(Debug, Clone, trustfall::provider::TrustfallEnumVertex)]
+enum V {
+    Variant(Value),
+}
+
+#[derive(Debug, Clone)]
+struct Value {
+    name: String,
+}
+impl Value {
+    fn initial(&self) -> String {
+        self.name
+            .chars()
+            .next()
+            .map(String::from)
+            .unwrap_or_default()
+    }
+}
+
+impl trustfall::provider::BasicAdapter<'static> for Adapter {
+    type Vertex = V;
+    fn resolve_starting_vertices(
+        &mut self,
+        _edge_name: &str,
+        _parameters: &trustfall::provider::EdgeParameters,
+    ) -> trustfall::provider::VertexIterator<'static, Self::Vertex> {
+        Box::new(std::iter::once(V::Variant(Value {
+            name: String::from("Berit"),
+        })))
+    }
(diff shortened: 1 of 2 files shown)


obi1kenobi:
Very nice!

Perhaps add an integration test to the `trustfall` crate? Integration tests (`tests/` directory instead of `src/`) compile as separate crates so it's a good place to make sure that `$crate` does what we'd expect even when the macro is re-exported.

ginger51011:
Alright, so I decided to import nothing to make it easier to ensure `FieldValue` was not imported, which does make the code utterly outrageous to look at. What do you think?

trustfall/tests/helper_macros.rs
```diff
@@ -0,0 +1,130 @@
...
+        _contexts: trustfall::provider::ContextIterator<'static, Self::Vertex>,
+        _type_name: &str,
+        _edge_name: &str,
+        _parameters: &trustfall::provider::EdgeParameters,
+    ) -> trustfall::provider::ContextOutcomeIterator<
+        'static,
+        Self::Vertex,
+        trustfall::provider::VertexIterator<'static, Self::Vertex>,
+    > {
+        todo!("schema should not contain neighbors: Berit likes it that way")
```
obi1kenobi, line 67: haha who is Berit? 😂

obi1kenobi (approved):

obi1kenobi:
Thanks for the fix! ❤️

trustfall/tests/helper_macros.rs
ginger51011: Extremely specific south Swedish internet sensation :smiley:

obi1kenobi: Amazing!

PR #288, opened 2023-05-17
Debug print query parse error

u9g:
Gives much more clarity to people why their query failed.

From:
image
To:
image

### DIFF
--- a/trustfall_wasm/src/lib.rs
+++ b/trustfall_wasm/src/lib.rs
@@ -74,7 +74,7 @@ pub fn execute_query(
     // TODO: add a proper error type
     let args = from_js_args(args)?;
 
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Thanks for flagging this! The underlying issue is that the parse error just hides the inner error's text instead of printing it. I've tweaked this in #289 so that the new error text here will read `Found a type coercion with sibling fields. Please move those fields inside the type coercion.`

In #289 I've also fixed a couple more error types that had the same problem, so you won't see "Query failed to parse" and no other info anymore. The error messages aren't perfect yet (I need to figure out how to include line number info in the error message) but they are certainly better now than they were before. Thanks for the help!

PR #289, opened 2023-05-17
Ensure we always output the inner error message.

obi1kenobi:
Fix the error pointed out in #288.


obi1kenobi:
image
It still needs a line number but it's a start!

PR #298, opened 2023-06-01
Replace `lazy_static` with `once_cell`

nvzqz:
The ecosystem has generally accepted that the approach used in `once_cell` is a better way to lazily initialize statics.

This change also makes it easier to later use [`std::sync::LazyLock`](https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html). See https://github.com/rust-lang/rust/issues/109736.


obi1kenobi (approved):
I haven't looked at some of this code in ages. It's always embarrassing to look back at old code and be like "what was I thinking, did I not know anything..?" 😅

Thanks!

nvzqz:
FWIW, I don't think `once_cell` was in common use or at 1.0 back when this code was first written. Also, many projects haven't yet moved past `lazy_static`.

PR #300, opened 2023-06-03
Refactor materializeWebsite in browser_based_querying

u9g:
This makes it much simpler to add new materializers based on urls, and simplifies the existing logic.

The various other changes are from prettier linting

### DIFF
--- a/experiments/browser_based_querying/src/hackernews/adapter.ts
+++ b/experiments/browser_based_querying/src/hackernews/adapter.ts
@@ -10,7 +10,7 @@ import {
   initialize,
   executeQuery,
 } from '../../www2/trustfall_wasm';
-import debug from "../utils/debug";
+import debug from '../utils/debug';
 import {
   getAskStories,
   getBestItems,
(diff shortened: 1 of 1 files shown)


experiments/browser_based_querying/src/hackernews/adapter.ts
```diff
@@ -71,25 +71,17 @@ const _itemPattern = /^https:\/\/news\.ycombinator\.com\/item\?id=(\d+)$/;
 const _userPattern = /^https:\/\/news\.ycombinator\.com\/user\?id=(.+)$/;
 
 function materializeWebsite(fetchPort: MessagePort, url: string): Vertex | null {
-  const itemMatch = url.match(_itemPattern);
-  let ret: any
-  if (itemMatch) {
+  let matcher: RegExpMatchArray | null = null;
+  let ret: any | null = { __typename: 'Website' };
```
obi1kenobi, line 75: Can `ret` ever be `null` here?

Can `materializeItem` or `materializeUser` return `null`, e.g. if the item or user have been deleted? If so, `ret.url` on line 83 might be a problem.

It seems that either the declared type here or the assignment on line 83 is buggy.

experiments/browser_based_querying/src/hackernews/adapter.ts
```diff
@@ -71,25 +71,17 @@ const _itemPattern = /^https:\/\/news\.ycombinator\.com\/item\?id=(\d+)$/;
...
+  if ((matcher = url.match(_itemPattern))) {
```
obi1kenobi, line 76: I'm not that familiar with TypeScript, is the double parens `((` required here, or is it just style, or something else?

obi1kenobi (commented):
Thanks for cleaning this code up! Just a couple of minor questions largely because I'm not that familiar with TypeScript.

experiments/browser_based_querying/src/hackernews/adapter.ts
u9g: I think it's a prettier thing, coming from the rule: https://eslint.org/docs/latest/rules/no-cond-assign

experiments/browser_based_querying/src/hackernews/adapter.ts
u9g: I fixed this in my code in the next commit, but really the code is scattered with direct materializeUser's which don't add .url so the fix is to just make them all defer back to materializeWebsite rather than directly calling materializeUser so the resulting object will have .url too, that will all be changed the next pr though, so I'm not overly worried.

u9g: Actually this isn't a problem because of even more indirection in which the url property is only read on Webpage objects, so their .url is never needed anyway.

obi1kenobi: I'm in favor of simplifying. It sounds like all that indirection makes stuff not obvious right now so feel free to tweak it to make it more maintainable and easier to understand.

obi1kenobi (approved):

PR #301, opened 2023-06-03
Stop passing fetchport through to funcs

u9g:
This heavily simplifies any function that needs to fetch because they no longer have to worry about plumbing the fetch function down through many layers, when really none of them need it except for the final destination

### DIFF
--- a/experiments/browser_based_querying/src/hackernews/adapter.ts
+++ b/experiments/browser_based_querying/src/hackernews/adapter.ts
@@ -70,18 +70,18 @@ function* limitIterator<T>(iter: IterableIterator<T>, limit: number): IterableIt
 const _itemPattern = /^https:\/\/news\.ycombinator\.com\/item\?id=(\d+)$/;
 const _userPattern = /^https:\/\/news\.ycombinator\.com\/user\?id=(.+)$/;
 
-function materializeWebsite(fetchPort: MessagePort, url: string): Vertex | null {
+function materializeWebsite(url: string): Vertex | null {
   let matcher: RegExpMatchArray | null = null;
   let ret: { url: string; __typename: string } | null = null;
   if ((matcher = url.match(_itemPattern))) {
     // This is an item.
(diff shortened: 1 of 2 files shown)


obi1kenobi:
I don't think this is a good change. Yes there's a bit of plumbing to pass the fetchport around, but that plumbing makes it obvious how the data is getting from A to B. Replacing it with global state saves a few characters' worth of typing at the expense of a lot of visibility and testability of the code.

obi1kenobi:
While I understand it's a bit more difficult up front to do a bit more plumbing up front, please keep in mind the broader context here: that tiny bit of extra upfront work is the kind of thing that makes the project viable at all as (right now) a single-maintainer project.

Let me explain.

> If fetch doesn't require a constant, why should this.

Fetch is a function that doesn't require any prior state to be set up. It's safe to call at any time and would work no matter what.

This isn't the case here. Fetching here requires that the `MessagePort` be set up first, and no fetching can happen before that. You can think of the `fetchPort` argument as a "proof" that this setup has been completed prior to attempting to fetch.

When `fetchPort` isn't passed around, it's easy to write code that fails to account for this invariant and attempts to fetch data before the setup is completed. Such code would fail at runtime, whereas the equivalent bug would be caught statically at type-checking time in the current approach with `fetchPort` being required to be passed in.

Trustfall *must* optimize for not having bugs, because bugs are an *existential* risk to the project. It has tons of surface area: multiple language bindings, a compiler, a query interpreter, various adapter implementations, the playground, etc. Time-wise, I simply cannot afford to risk bugs popping up and requiring me to do lots of triage and debugging, or else I risk burnout and that means the project dies completely.

Between a tiny bit of extra work up front to make a type of bug less likely to happen at all, and a bit less work but at a risk of future bugs and debugging, for the sake of the long-term viability of the project I have to *always* pick the former.

I know that's not the answer you were hoping for but I hope you can see where I'm coming from here.

u9g:
Understandable

PR #310, opened 2023-06-10
Convert all vertexes to classes

u9g:
So far only properties are done hence the draft status, but it works for properties now.

### DIFF
--- a/experiments/browser_based_querying/src/hackernews/adapter.ts
+++ b/experiments/browser_based_querying/src/hackernews/adapter.ts
@@ -26,6 +26,7 @@ import {
   materializeUser,
 } from './utils';
 import HN_SCHEMA from './schema.graphql';
+import { Webpage } from './data/Webpage';
 
 initialize(); // Trustfall query system init.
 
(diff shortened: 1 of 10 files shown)


experiments/browser_based_querying/src/hackernews/adapter.ts
```diff
@@ -128,7 +129,7 @@ function* linksInAboutPage(
     }
 
     const aboutPlain = extractPlainTextFromHnMarkup(aboutHtml);
-    const matches2 = aboutPlain.matchAll(/http[s]?:\/\/[^ \n\t]*[^ \n\t\.);,\]}]/g);
```
obi1kenobi, line 131: Are you sure about this change?

u9g: the difference is we don't escape a . in a [] which is fine (and is a prettier error). I can revert if you want. https://eslint.org/docs/latest/rules/no-useless-escape

obi1kenobi: Hmm interesting, I thought `[]` accepted character groups like `.` 🤔 

Any idea why our linters didn't already warn on this? We have a JS/TS lint job running in CI.

(Side note: as a norm in open-source development, it's usually better to fix things like these in a separate PR. It's easier to merge and offers the chance to explain why the change is being made. When you "slip it in" into some other PR, it risks looking accidental or even sloppy.)

u9g: Understandable

u9g:
We decided to go in a different direction of data-driven design rather than oop

PR #311, opened 2023-06-11
Adds lowest_common_denom method

u9g:
Enables creating data-driven resolvers. With this, you can do `schema.lowest_common_denom("Story", "url")` and get back `Webpage`, which enables you to match on the lowest common denominator that has this field, rather than match on every field that implements the url field such as `Story`, `Comment`, `Webpage`, etc. (These types were taken from the hackernews example)


u9g:
Won't work

PR #313, opened 2023-06-12
Replace HRTBs with `DeserializeOwned`

nvzqz:
Replaces `for<'de> Deserialize<'de>` [higher-rank trait bounds (HRTBs)](https://doc.rust-lang.org/nomicon/hrtb.html) with the much simpler [`DeserializeOwned`](https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html) helper trait.

This change has no effect on semantics or compatibility.


obi1kenobi:
Thanks!

PR #314, opened 2023-06-12
Simplify definitions and impls of `Vertex`-containing types

nvzqz:
By not requiring so many bounds on `Vertex` at type definition time, it becomes easier to modify this code.


nvzqz:
I kept the trait bounds on `DataContext::new`, so that shouldn't be an issue 🙂

https://github.com/obi1kenobi/trustfall/blob/7b36bfac678996b6e7143d3598c656876f69a5cc/trustfall_core/src/interpreter/mod.rs#L183-L195

PR #339, opened 2023-07-03
Add optimization to VertexType entrypoint in schema adapter

u9g:


trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -214,15 +215,34 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
                 let root_query_type = self.schema.query_type_name();
-                Box::new(self.schema.vertex_types.values().filter_map(move |v| {
-                    (v.name.node != root_query_type)
-                        .then(|| SchemaVertex::VertexType(VertexType::new(v)))
-                }))
+
+                if let Some(crate::interpreter::CandidateValue::Single(FieldValue::String(str))) =
+                    resolve_info.statically_required_property("name")
+                {
+                    let str: Rc<str> = str.as_str().into();
```
obi1kenobi, line 227: Do we need this to be `Rc<str>`? It seems to only be used via `.as_ref()` which produces the same thing as `str.as_str()` would.

obi1kenobi (commented):
One minor nit, otherwise lgtm! 🚀

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -214,15 +215,34 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
     ) -> VertexIterator<'a, Self::Vertex> {
         match edge_name.as_ref() {
             "VertexType" => {
                 let root_query_type = self.schema.query_type_name();
-                Box::new(self.schema.vertex_types.values().filter_map(move |v| {
-                    (v.name.node != root_query_type)
-                        .then(|| SchemaVertex::VertexType(VertexType::new(v)))
-                }))
+
+                if let Some(crate::interpreter::CandidateValue::Single(FieldValue::String(str))) =
```
obi1kenobi, line 224: Can we handle `CandidateValue::Multiple` here?

Otherwise queries with `name @filter(op: "=", ...)` will be fast, but queries with `name @filter(op: "one_of", ...)` will be slow.

obi1kenobi (commented):
One other idea here!

trustfall_core/src/schema/adapter/mod.rs
u9g: You're right, fixed!

trustfall_core/src/schema/adapter/mod.rs
u9g: Done!

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -214,15 +215,62 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
+        let candidate_value = resolve_info.statically_required_property("name").clone();
         match edge_name.as_ref() {
             "VertexType" => {
                 let root_query_type = self.schema.query_type_name();
-                Box::new(self.schema.vertex_types.values().filter_map(move |v| {
-                    (v.name.node != root_query_type)
-                        .then(|| SchemaVertex::VertexType(VertexType::new(v)))
-                }))
+
+                if let Some(crate::interpreter::CandidateValue::Single(FieldValue::String(
```
obi1kenobi, line 225: Can you add a `use` for `CandidateValue` so that these lines don't end up being so long?

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -214,15 +215,62 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
+                    Box::new(
+                        possibilities_as_owned_strings
+                            .into_iter()
+                            .filter_map(move |wanted| {
+                                vertex_types
+                                    .get(wanted.as_str())
+                                    .filter(move |v| v.name.node != root_query_type)
+                                    .map(|exact_wanted| {
+                                        SchemaVertex::VertexType(VertexType::new(exact_wanted))
+                                    })
```
obi1kenobi, line 265: Try `.and_then()` to combine these two.

trustfall_core/src/schema/adapter/tests.rs
```diff
@@ -42,6 +42,53 @@ fn check_vertex_type_properties() {
     assert_eq!(expected_rows, rows);
 }
 
+#[test]
+fn check_vertex_type_properties_using_one_of() {
+    let query = r#"
+{
+    VertexType {
+        name @filter(op: "one_of", value: ["$name"])
```
obi1kenobi, line 50: Could you make this output the vertex type name as well?
```suggestion
        name @filter(op: "one_of", value: ["$name"]) @output
```

Otherwise the fact that `"property" => "name"` appears twice seems like it could be a bug, and it isn't obvious why it's correct.

obi1kenobi (commented):
Just a few minor nitpicks, should be good to go after these tweaks.

trustfall_core/src/schema/adapter/mod.rs
u9g: Done!

trustfall_core/src/schema/adapter/mod.rs
u9g: Done!

Issue #356, opened 2023-07-07
Schema::new unwrapping None


u9g:
Schema in question: https://gist.github.com/u9g/7194a28bfd9af3e201ac10e269e69678

Error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: HashMapOccupiedError { entry: OccupiedEntry { key: ("JSXElementAST", "span"), value: FieldDefinition { description: None, name: Positioned { pos: Pos(301:3), node: Name("span") }, arguments: [], ty: Positioned { pos: Pos(301:9), node: Type { base: Named(Name("Span")), nullable: false } }, directives: [] }, .. }, value: FieldDefinition { description: None, name: Positioned { pos: Pos(303:3), node: Name("span") }, arguments: [], ty: Positioned { pos: Pos(303:9), node: Type { base: Named(Name("Span")), nullable: false } }, directives: [] } }', trustfall_core\src\schema\mod.rs:172:34
```

obi1kenobi:
The schema error is that `span` appears twice in the `JSXElementAST` type's list of fields.

I had hoped that the GraphQL parser we're using would have caught and reported this error already, but apparently not... 😒

u9g:
Yup, that was it, thanks for such a quick reply!

PR #373, opened 2023-07-12
add InvalidSchemaError::DuplicateFieldDefinition

u9g:
closes https://github.com/obi1kenobi/trustfall/issues/356

### DIFF
--- a/trustfall_core/src/schema/error.rs
+++ b/trustfall_core/src/schema/error.rs
@@ -126,7 +126,7 @@ pub enum InvalidSchemaError {
     )]
     ImplementingNonInterface(String, String),
 
-    #[error("Type \"{0}\" defines the field \"{1}\" multiple times.")]
+    #[error("Type or Interface \"{0}\" defines the field \"{1}\" multiple times.")]
     DuplicateFieldDefinition(String, String),
 }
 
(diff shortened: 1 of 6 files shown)


obi1kenobi:
Very nice!

Can you add a test case for this? You can add a new test schema file in `trustfall_core/test_data/tests/schema_errors/` that triggers this error, then run `scripts/create_schema_error_data.sh <schema_file>` and make sure the newly generated schema error output is correct.

u9g:
> Very nice!
> 
> Can you add a test case for this? You can add a new test schema file in `trustfall_core/test_data/tests/schema_errors/` that triggers this error, then run `scripts/create_schema_error_data.sh <schema_file>` and make sure the newly generated schema error output is correct.

Done!

trustfall_core/src/schema/error.rs
```diff
@@ -125,6 +125,9 @@ pub enum InvalidSchemaError {
         Only interfaces can be implemented by other types."
     )]
     ImplementingNonInterface(String, String),
+
+    #[error("Type or Interface \"{0}\" defines the field \"{1}\" multiple times.")]
```
obi1kenobi, line 129:
```suggestion
    #[error("Type \"{0}\" defines the field \"{1}\" multiple times.")]
```
GraphQL terminology (and by extension Trustfall terminology) is a mess here, but all vertex types are types, and `type` in GraphQL is usually called "object type" which is a term I don't like but I don't have a better one. In any case, including both terms here is probably overall more confusing, so let's stick to the simpler case.

obi1kenobi (approved):

PR #387, opened 2023-07-20
Add suppressions for a couple of new clippy lints.

obi1kenobi:


obi1kenobi:
This lint is still unstable and only in 1.73's clippy, so I'll reopen this PR if/when it becomes relevant again.

PR #393, opened 2023-07-22
Run-length + custom dictionary compression for playground query URLs.

obi1kenobi:
Losslessly encodes query and variables data to cut length to approx. ~55% of prior size. Bigger queries benefit more.

Important bits of the design:
- Percent-encoding in URLs is massively expensive at 3x bloat, avoid it whenever possible.
- `-` and `*` are URL-safe characters that are never percent-encoded by browsers (not even when copying out of the address bar). They are also ~never used in GraphQL (only in comments or string literals), so it's cost-effective to use them as part of escape sequences.
- Queries tend to have a lot of spaces often preceded by `}\n` or `{\n`, so use run-length encoding on the spaces.
- Use a custom dictionary for common fixed syntactic elements like directives: `@output` becomes `*o` etc.
- Single spaces are common, and we want continuous text (e.g. in comments) to appear nice in the URL and not be too expensive, so encode single spaces as `-`.
- Don't assume people write perfectly-formatted queries. For example, if someone misses a space before a `{`, we don't want that to suddenly explode their URL size.

### DIFF
--- a/experiments/browser_based_querying/src/urlCompression.ts
+++ b/experiments/browser_based_querying/src/urlCompression.ts
@@ -1,0 +1,361 @@
+// Decoding tables:
+// "-" => " "
+// "--_0" thru "--_9", then "--_a" thru "--_z", then "--_A" thru "--_Z" => 2+ spaces, run-length encoded
+// "--0" thru "--9", then "--a" thru "--z", then "--A" thru "--Z" => "{\n" followed by 2+ spaces, run-length encoded
+// "--*0" thru "--*9", then "--*a" thru "--*z", then "--*A" thru "--*Z" => "}\n" followed by 2+ spaces, run-length encoded
+// N.B.: "---" is a not a legal codeword because it means "space followed by '--' escape sequence"
+//
+// "*l" => "\n"
+// "*L" => ","
+// "**" => '"'
+// "*-" => "-"
+// "*c" => "["
+// "*j" => "]"
+// "*C" => "{"
+// "*J" => "}"
+// "*B" => ":"
+// "*1" => "!"
+// "*2" => "@"
+// "*3" => "#"
+// "*4" => "$"
+// "*5" => "%"
+// "*8" => "*"
+// "*9" => "("
+// "*0" => ")"
+// "*g" => "<"
+// "*G" => ">"
+// "*e" => "="
+// "*o" => "@output"
+// "*O" => "@optional"
+// "*t" => "@tag"
+// "*f" => "@filter"
+// "*T" => "@transform"
+// "*F" => "@fold"
+// "*r" => "@recurse"
+// "*E" => "... on "
+// "*n" => "name:"
+// "*d" => "depth:"
+// "*v" => "value:"
+// "*p" => "op:"
+
+function invertAndStripPrefix(data: Record<string, string>, prefix: string): Record<string, string> {
+  const result: Record<string, string> = {};
+  for (const key in data) {
+    let value = data[key];
+    if (value.startsWith(prefix)) {
+      value = value.substring(prefix.length);
+    }
+    result[value] = key;
+  }
+  return result;
+}
+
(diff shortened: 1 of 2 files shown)


obi1kenobi:
Still needs the wiring for:
- decoding URL data based on the format value
- upgrading `f=1` URLs to `f=2` compressed URLs

obi1kenobi:
Example URL parameters for the "structs importable by multiple paths" query from the playground, before (613 chars) vs after (345 chars):
```
?f=1&q=IyBTdHJ1Y3RzIGltcG9ydGFibGUgZnJvbSB0aGUgY3JhdGUgdmlhIG11bHRpcGxlIHBhdGhzCnF1ZXJ5IHsKICBDcmF0ZSB7CiAgICBpdGVtIHsKICAgICAgLi4uIG9uIFN0cnVjdCB7CiAgICAgICAgbmFtZSBAb3V0cHV0CgogICAgICAgIGltcG9ydGFibGVfcGF0aCBAZm9sZAogICAgICAgICAgICAgICAgICAgICAgICBAdHJhbnNmb3JtKG9wOiAiY291bnQiKQogICAgICAgICAgICAgICAgICAgICAgICBAZmlsdGVyKG9wOiAiPj0iLCB2YWx1ZTogWyIkcGF0aF9jb3VudCJdKQogICAgICAgIHsKICAgICAgICAgIHBhdGggQG91dHB1dAogICAgICAgIH0KCiAgICAgICAgc3BhbiB7CiAgICAgICAgICBmaWxlbmFtZSBAb3V0cHV0CiAgICAgICAgICBmaXJzdF9saW5lOiBiZWdpbl9saW5lIEBvdXRwdXQKICAgICAgICB9CiAgICAgIH0KICAgIH0KICB9Cn0%3D&v=ewogICJwYXRoX2NvdW50IjogMgp9
?f=2&q=*3-Structs-importable-from-the-crate-via-multiple-paths*lquery---0Crate---2item---4*E-Struct---6name-*o*l*l--_6importable_path-*F*l--_m*T*9*p-**count***0*l--_m*f*9*p-***G*e***L-*v-*c***4path_count***j*0*l--_6--8path-*o*l--_6*J*l*l--_6span---8filename-*o*l--_8first_line*B-begin_line-*o*l--_6--*4--*2--*0*J*l*J&v=--0**path_count***B-2*l*J
```

obi1kenobi:
@nvzqz heads up in case you're interested, since I think you also gave this problem some thought!

Also @paavanb thanks for adding that `f=1` parameter, I remember we talked about it way back and it's gonna make this so much easier now :)

experiments/browser_based_querying/src/urlCompression.ts
```diff
@@ -0,0 +1,361 @@
...
+  '@fold': '*F',
+  '@recurse': '*r',
+}
+
+const SPECIAL_CHAR_REPLACEMENTS: Record<string, string> = {
+  '"': '**',
+  '-': '*-',
+  '[': '*c',
+  ']': '*j',
+  '{': '*C',
```
nvzqz, line 68: We can further compress the common space-prefixed `{` if we have either:
- An encoding token that indicates "remove the following space". This would also improve the compression of poorly-formatted `@` tokens in my other suggestion
- Two separate `{` and ` {` encoding tokens

obi1kenobi:
I generally avoided compression opportunities that rely on good formatting and/or sacrifice manual URL readability in order to get small gains. I feel like the tradeoff there is not worth it, because it will:
- hurt users that aren't very good at Trustfall (e.g. don't write pretty-formatted queries by default)
- make their pain less visible to us, because their URLs will be longer which makes them more likely to be cut off when sharing over Zoom / Google Meet
- make their URLs harder to read and debug if they happen to report the problem to us

in the current encoding scheme, seeing a `-` means there's _always_ a space, at least one and possibly more. I think it's not worth sacrificing these properties to squeeze the last few bytes' worth of compression out of this.

Chatted directly & resolved.

PR #403, opened 2023-07-25
Make `test_field_value_into` nicer to work with

nvzqz:
This change reduces verbosity and makes any panics use the location for the input data.


obi1kenobi:
:+1:

PR #404, opened 2023-07-25
Format with max small heuristics

nvzqz:
This makes formatting take more advantage of the same line, reducing the number of lines used.

I find that this tends to improve readability when:
- Passing closures to chained functions
- Pattern matching is in a heavily indented context
- There is opportunity for vertical alignment in neighboring code


obi1kenobi:
This looks neat. If you don't mind re-running `cargo fmt` again to resolve the merge conflicts, I'll merge it.

PR #413, opened 2023-07-28
add schema starting edge to schemaadapter

u9g:


trustfall_core/src/schema/adapter/schema.graphql
```diff
@@ -1,150 +1,167 @@
...
+
+  """
+  The entry point edges at which querying may begin.
+
+  Corresponds to the valid edge names for the `resolve_starting_vertices()`
+  method for adapters over this schema.
+  """
+  Entrypoint: [Edge!]!
+
+  Schema: Schema!
```
u9g: didn't know what the comment should be here so I left it blank but we probably should add one

trustfall_core/src/schema/adapter/schema.graphql
```diff
@@ -18,6 +18,25 @@ type RootSchemaQuery {
     method for adapters over this schema.
     """
     Entrypoint: [Edge!]!
+    
+    Schema: Schema!
```
u9g, line 22: didn't know what the comment should be here so I left it blank but we probably should add one

trustfall_core/src/schema/adapter/tests.rs
```diff
@@ -99,6 +99,55 @@ fn check_vertex_type_properties_using_one_of() {
...
+    }
+    .into();
+    let adapter = Arc::new(SchemaAdapter::new(&SCHEMA));
+
+    let indexed = crate::frontend::parse(&SCHEMA, query).expect("not a valid query");
+    let mut rows: Vec<_> = crate::interpreter::execution::interpret_ir(adapter, indexed, args)
+        .expect("execution error")
+        .collect();
+
+    rows.sort_by(|a, b| a["name"].partial_cmp(&b["name"]).expect("to be comparable"));
```
u9g, line 127: this is actually a problem in the test above aswell, we compared based on "property" rather than "name" so the sort is unstable, it needs to be fixed there too in a future pr

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -400,6 +408,23 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
                 _ => unreachable!("unexpected edge name on type {type_name}: {edge_name}"),
             },
+            "Schema" => match edge_name.as_ref() {
+                "vertex_type" => {
+                    let candidate_destination = resolve_info.destination();
+
+                    let candidate_value_for_name =
+                        candidate_destination.statically_required_property("name");
+
+                    let mut iter = Some(self.vertex_type_iter(candidate_value_for_name));
```
u9g, line 418: putting this into a Some() to .take it so we don't have to worry about lifetime concerns about calling into `self.` in the callback. not my best work, but atleast it works :)

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -400,6 +408,23 @@ impl<'a> crate::interpreter::Adapter<'a> for SchemaAdapter<'a> {
...
+
+                    resolve_neighbors_with(contexts, move |_| iter.take().unwrap())
+                }
+                "entrypoint" => {
+                    let mut iter = Some(self.entrypoints_iter());
```
u9g, line 423: here too

u9g (commented):
some thoughts

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -73,6 +73,61 @@ impl<'a> SchemaAdapter<'a> {
     pub fn schema_text() -> &'static str {
         include_str!("./schema.graphql")
     }
+
+    pub fn vertex_type_iter(
```
obi1kenobi, line 77: This is definitely not a fit to be a `pub fn` (we don't want to commit to publicly maintaining it forever) and probably also shouldn't be a method on the adapter either. It can just as easily be an independent private function that takes a `&Schema` argument.

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -73,6 +73,61 @@ impl<'a> SchemaAdapter<'a> {
...
+            if let Some(exact_wanted) = self
+                .schema
+                .vertex_types
+                .get(name_wanted)
+                .filter(move |v| v.name.node != root_query_type)
+            {
+                Box::new(std::iter::once(SchemaVertex::VertexType(VertexType::new(exact_wanted))))
+            } else {
+                Box::new(std::iter::empty())
+            }
```
obi1kenobi, line 96: remember to `.into_iter()` on the option

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -73,6 +73,61 @@ impl<'a> SchemaAdapter<'a> {
...
+        } else if let Some(CandidateValue::Multiple(possibilities)) = candidate_value_for_name {
+            let possibilities_as_owned_strings = possibilities
+                .iter()
+                .map(|el| el.as_str().expect("for name possibility to be a string").to_string())
```
obi1kenobi, line 100: Oh with the new `FieldValue` representation, there's a better option here: we could add an `as_arc_str() -> Option<Arc<str>>` method on `FieldValue` and save on the deep-clone of the string and probably _also_ on the vector as well.

Wanna open a PR for that, or should I?

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -73,6 +73,61 @@ impl<'a> SchemaAdapter<'a> {
...
+        } else {
+            Box::new(self.schema.vertex_types.values().filter_map(move |v| {
+                (v.name.node != root_query_type)
+                    .then(|| SchemaVertex::VertexType(VertexType::new(v)))
+            }))
+        }
+    }
+
+    pub fn entrypoints_iter(&self) -> VertexIterator<'a, SchemaVertex<'a>> {
+        Box::new(Box::new(
```
obi1kenobi, line 123: why are we double-boxing this?

trustfall_core/src/schema/adapter/mod.rs
```diff
@@ -73,6 +73,61 @@ impl<'a> SchemaAdapter<'a> {
...
+            }))
+        } else {
+            Box::new(self.schema.vertex_types.values().filter_map(move |v| {
+                (v.name.node != root_query_type)
+                    .then(|| SchemaVertex::VertexType(VertexType::new(v)))
+            }))
+        }
+    }
+
+    pub fn entrypoints_iter(&self) -> VertexIterator<'a, SchemaVertex<'a>> {
```
obi1kenobi, line 122: also not a fit for a `pub fn`, same reasoning as above

trustfall_core/src/schema/adapter/mod.rs
obi1kenobi: This is extremely sketchy. I register my complaint, as predicted.

I think if you make this be a free function instead of a method, that will solve the lifetime problem: the schema reference is `Copy` and lives long enough, and it won't matter whether `self` lives as long as the iterator or not.

trustfall_core/src/schema/adapter/mod.rs
obi1kenobi: Same solution as in the other place.

trustfall_core/src/schema/adapter/schema.graphql
obi1kenobi: How about: `Information about the schema itself.` or something along those lines?

trustfall_core/src/schema/adapter/mod.rs
obi1kenobi: Merged that PR: #414

trustfall_core/src/schema/adapter/mod.rs
u9g: so this is actually right

trustfall_core/src/schema/adapter/mod.rs
u9g: Sounds good

trustfall_core/src/schema/adapter/schema.graphql
u9g: Sounds great

trustfall_core/src/schema/adapter/schema.graphql
```diff
@@ -18,6 +18,28 @@ type RootSchemaQuery {
...
+type Schema {
+    """
+    All types of vertices present in this schema.
+    """
+    vertex_type: [VertexType!]!
+
+    """
+    The entry point edges at which querying may begin.
+    Corresponds to the valid edge names for the `resolve_starting_vertices()`
+    method for adapters over this schema.
```
obi1kenobi, line 40:
```suggestion

    Corresponds to the valid edge names for the `resolve_starting_vertices()`
    method for adapters over this schema.
```
to match the formatting near lines 14-20

trustfall_core/src/schema/adapter/tests.rs
obi1kenobi: Could you fix it here by using the `TryIntoStruct` trait into a struct with a `#[derive(PartialOrd, Ord)]` which will then provide a guaranteed sort order?

We should apply the same fix in the other test as well in a subsequent PR.

obi1kenobi (commented):
Looks good, just two tiny nitpicks.

obi1kenobi (approved):
Awesome, ship it 🚀

PR #419, opened 2023-07-31
Run doctests in CI and fix failures.

obi1kenobi:
Unfortunately, adding `--all-targets` to `cargo test` causes doctests to
be de-selected: https://github.com/rust-lang/cargo/issues/6669


obi1kenobi:
h/t @u9g for finding this issue

PR #421, opened 2023-08-02
Add new test queries

u9g:


trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_less_than.graphql.ron
```diff
@@ -0,0 +1,18 @@
...
+
+            primeFactor @fold @transform(op: "count") @filter(op: "<", value: ["$min_primes"]) {
+                __typename
+            }
+        }
+    }
+}"#,
+    arguments: {
+        "min_primes": Uint64(2)
+    },
```
obi1kenobi, line 17: Can you tweak this query so it returns at least one result? That way we won't have to stare too hard at the trace if we have a bug that causes all results to disappear.

Perhaps just make the entrypoint have `max: 32` which should return just 32 (since 31 is prime).

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_greater_than.graphql.ron
```diff
@@ -0,0 +1,19 @@
...
+TestGraphQLQuery (
+    schema_name: "numbers",
+    query: r#"
+{
+    Number(min: 30, max: 30) {
+        ... on Composite {
+            value @output
+
+            primeFactor @fold @transform(op: "count") @filter(op: ">", value: ["$min_primes"]) {
+                __typename @output
```
obi1kenobi, line 11: Consider outputting a more interesting field than `__typename`, so that we get different results and we get a bit more testing power. For example, output `prime: value`.

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_less_than.graphql.ron
```diff
@@ -0,0 +1,18 @@
...
+            primeFactor @fold @transform(op: "count") @filter(op: "<", value: ["$min_primes"]) {
+                __typename @output
+            }
+        }
+    }
+}"#,
+    arguments: {
+        "min_primes": Uint64(2)
+    },
+)
```
obi1kenobi, line 18: Similar to earlier, tweak to get some results.

obi1kenobi (commented):
Some suggestions on how to get more testing power out of the tests, otherwise happy to merge.

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_greater_than.graphql.ron
```diff
@@ -0,0 +1,22 @@
+TestGraphQLQuery (
+    // This @fold statement doesn't actually need to fully evaluate:
+    // it has no outputs, and the predicate can be determined to be satisfied
+    // with only a partial evaluation of the @fold.
```
obi1kenobi, line 4: I don't think this is true currently, actually. In order to be `> 2`, the fold needs to see three elements -- and there are only three prime factors, so this `@fold` essentially gets fully evaluated. (Yes, small technicality here, Trustfall doesn't have to _determine_ that is has fully evaluated it by verifying that it has drained its iterator, but it will have drained it all the same.)

Consider making the argument be `1` instead of `2`?

trustfall_core/test_data/tests/valid_queries/fold_with_no_inner_output_with_count_output.graphql.ron
```diff
@@ -0,0 +1,28 @@
+TestGraphQLQuery (
+    schema_name: "numbers",
+    // This @fold statement doesn't actually need to fully evaluate:
+    // it has no outputs, and the predicate can be determined to be satisfied
+    // with only a partial evaluation of the @fold.
+    //
+    // The number 30 is composite and has 3 prime factors: 2, 3, 5.
+    // This test checks whether the evaluation of the @fold is stopped early
+    // after two of those three primes are produced (satisfying the predicate).
```
obi1kenobi, line 9: I think this is not true — in fact the opposite is true. The `@output` on the count requires us to fully evaluate the `@fold`, or else we don't know what the count is.

Did you mean to put this comment in a different file?

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.graphql.ron
```diff
@@ -0,0 +1,22 @@
+TestGraphQLQuery (
+    schema_name: "numbers",
```
obi1kenobi, line 2: An explanatory comment here would go a long way toward explaining what this test case is about.

Also, consider a variation of this test case that would make it a bit more interesting: What if the largest number of `min_primes` is enough below the actual count that there's an opportunity to discard the result early when it's clear it won't ever match? Does our optimization kick in then? Do we already have a test case for this?

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_greater_than.graphql.ron
```diff
@@ -0,0 +1,19 @@
+
```
obi1kenobi, line 1:
```suggestion
```
try to catch this stuff in self-review

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_greater_than.graphql.ron
```diff
@@ -0,0 +1,19 @@
...
+            primeFactor @fold @transform(op: "count") @filter(op: ">", value: ["$min_primes"]) {
+                prime: value @output
+            }
+        }
+    }
+}"#,
+    arguments: {
+        "min_primes": Uint64(2)
+    },
```
obi1kenobi, line 18: `> 2` here also requires fully evaluating the `@fold` because it requires seeing a third element, and the fold only has 3 elements. Consider making `min_primes` be 1 instead of 2 to make this a test of the optimization's behavior.

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_less_than.graphql.ron
```diff
@@ -0,0 +1,18 @@
...
+            primeFactor @fold @transform(op: "count") @filter(op: "<", value: ["$min_primes"]) {
+                prime: value @output
+            }
+        }
+    }
+}"#,
+    arguments: {
+        "min_primes": Uint64(2)
+    },
+)
```
obi1kenobi, line 18: Not sure what this test case is testing? Also it seems to have no results because the `< 2` predicate on the fold filters everything out.

obi1kenobi (commented):
Consider (either here or in a separate PR) also including a test case where there is both a `>=` and a `<=` filter on the count and there's no outputs in the fold. It would make sure the new optimization doesn't kick in in situations where the count can be observed indirectly via a filter that could flip from passing to not passing as the count grows.

Two variations worth testing here:
- when we count to a number that clears the `>=`, and then hit the end of the `@fold` iterator before we hit the `<=` number, the result is returned
- when we exceed the `<=` number, we stop counting immediately regardless of everything else

Another variation of that test case: have both a `>=` and a `one_of` filter. In this case, we need the exact count always, because that's the only way to match the `one_of` filter.

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_greater_than.graphql.ron
u9g: Yeah I can change it, but I actually wrote this comment after testing with my patch, it does result in a diff after my patch, so it was partially skipped. I'll change it though, no problem.

trustfall_core/test_data/tests/valid_queries/fold_with_no_inner_output_with_count_output.graphql.ron
u9g: Yeah, it's from an earlier test that I copied, not meant to be here after I added the `@output`. I've fixed it :)

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_greater_than.graphql.ron
u9g: Oh oops, thanks!

trustfall_core/test_data/tests/valid_queries/fold_with_outputs_transform_and_filter_less_than.graphql.ron
u9g: I made the same test case for > and <, I didn't think it would be a problem to add another behavioral test case to ensure no breaks in existing behavior.

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.graphql.ron
u9g: This test case actually doesn't fit anywhere since we don't at all touch one_of in the optimization pr I'm about to do

u9g: This test case has been repurposed to that^

trustfall_core/test_data/tests/valid_queries/fold_with_no_inner_output_with_count_output.graphql.ron
```diff
@@ -0,0 +1,27 @@
+TestGraphQLQuery (
+    schema_name: "numbers",
+    // This @fold statement needs to evaluate because the @output forces us to observe the
+    // full iterator in order to calculate length.
+    //
+    // The number 30 is composite and has 3 prime factors: 2, 3, 5.
+    // This test checks whether the evaluation of the @fold is stopped early
+    // after two of those three primes are produced (satisfying the predicate).
```
obi1kenobi, line 8:
```suggestion
    // The number 30 is composite and has 3 prime factors: 2, 3, 5.
    // This test ensures that we don't incorrectly stop the evaluation of the @fold
    // early, after producing only two of those primes thereby satisfying the predicate.
```
just edit for clarity, to make it obvious what the _expected_ behavior is

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_greater_than.graphql.ron
obi1kenobi: With `> 1` as the predicate I think this is fine.

I suspect the change you were previously observing was just that `InputIteratorExhausted` trace operations were being elided, that's the technicality I was referring to. But I like this more explicit version better.

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.graphql.ron
```diff
@@ -0,0 +1,22 @@
...
+
+            primeFactor @fold @transform(op: "count") @filter(op: "one_of", value: ["$min_primes_count"]) {
+                __typename
+            }
+        }
+    }
+}"#,
+    arguments: {
+        "min_primes_count": List([
+            Uint64(1),
```
obi1kenobi, line 19:
```suggestion
            Uint64(1),
            Uint64(0),
```
to ensure our code tests with the max value, not an arbitrary value and not the _last_ value in the list

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.graphql-parsed.ron
```diff
@@ -0,0 +1,98 @@
...
+              ),
+            ],
+          )),
+        )),
+      ],
+    ),
+  ),
+  arguments: {
+    "min_primes_count": List([
+      Uint64(1),
```
obi1kenobi, line 95:
```suggestion
      Uint64(1),
      Uint64(0),
```
to match the suggestion below

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.ir.ron
```diff
@@ -0,0 +1,60 @@
...
+        ),
+      },
+    ),
+    variables: {
+      "min_primes_count": "[Int!]!",
+    },
+  ),
+  arguments: {
+    "min_primes_count": List([
+      Uint64(1),
```
obi1kenobi, line 57:
```suggestion
      Uint64(1),
      Uint64(0),
```

trustfall_core/test_data/tests/valid_queries/fold_with_no_outputs_transform_and_filter_one_of.trace.ron
```diff
@@ -0,0 +1,223 @@
...
+          ),
+        },
+      ),
+      variables: {
+        "min_primes_count": "[Int!]!",
+      },
+    ),
+    arguments: {
+      "min_primes_count": List([
+        Uint64(1),
```
obi1kenobi, line 219:
```suggestion
        Uint64(1),
        Uint64(0),
```

PR #423, opened 2023-08-02
add min_fold_count_limit optimization

u9g:
Stop expanding `@fold` early if the fold contains no `@output` fields and the count of the fold is not outputted, while also not invaliding other filters by skipping the rest of the iterator such as filtering for an upperbound. (`@filter(< 10)`)

### DIFF
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -341,7 +341,11 @@ fn get_min_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
                         .expect("for field value to be coercible to usize");
                 Some(variable_value.saturating_add(1))
             }
-            _ => None,
+            // This optimization is only valid if we know that every
+            // filter applied to the folded element count is a comparison
+            // that we can determine the number of elements needed to satisfy
+            // the filter.
+            _ => return None,
         };
 
         match (result, next_limit) {
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -364,6 +364,7 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
     min_fold_count_limit: &Option<usize>,
     no_outputs_in_fold: bool,
     has_output_on_fold_count: bool,
+    has_tag_on_fold_count: bool,
 ) -> Option<Vec<DataContext<Vertex>>> {
     // We do not apply the min_fold_count_limit optimization if we have an upper bound,
     // in the form of max_fold_count_limit, because if we have a required upperbound,
@@ -412,7 +413,9 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
             // When we do `iterator.take(*min_fold_count_limit)`, this statement takes at most
             // `min_fold_count_limit` elements, however if we have less the iterator can stop
             // early.
-            Some(min_fold_count_limit) if no_outputs_in_fold && !has_output_on_fold_count => {
+            Some(min_fold_count_limit)
+                if no_outputs_in_fold && !has_output_on_fold_count && !has_tag_on_fold_count =>
+            {
                 iterator.take(*min_fold_count_limit).collect()
             }
             // We weren't able to find any early-termination condition for materializing the fold,
@@ -515,6 +518,23 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
     let no_outputs_in_fold = fold.component.outputs.is_empty();
     let has_output_on_fold_count =
         fold.fold_specific_outputs.values().any(|x| *x == FoldSpecificFieldKind::Count);
+    let has_tag_on_fold_count = parent_component
+        .vertices
+        .iter()
+        .flat_map(|x| {
+            x.1.filters.iter().filter(|x| {
+                let Some(Argument::Tag(FieldRef::FoldSpecificField(tagged_fold_count))) = x.right()
+                else {
+                    return false;
+                };
+
+                tagged_fold_count.fold_root_vid == fold.to_vid
+                    && tagged_fold_count.fold_eid == fold.eid
+                    && tagged_fold_count.kind == FoldSpecificFieldKind::Count
+            })
+        })
+        .next()
+        .is_some();
     let moved_fold = fold.clone();
     let folded_iterator = edge_iterator.filter_map(move |(mut context, neighbors)| {
         let imported_tags = context.imported_tags.clone();
@@ -544,6 +564,7 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
                 &min_fold_size,
                 no_outputs_in_fold,
                 has_output_on_fold_count,
+                has_tag_on_fold_count,
             )?)
         } else {
             None
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -241,6 +241,27 @@ fn construct_outputs<'query, AdapterT: Adapter<'query>>(
     }))
 }
 
+/// Extracts numeric [`FieldValue`] into a `usize`, clamping negative numbers to 0.
+/// Returns `None` on `FieldValue::Null`, and panics otherwise.
+fn usize_from_field_value(field_value: &FieldValue) -> Option<usize> {
+    match field_value {
+        FieldValue::Int64(num) => {
+            Some(usize::try_from(*num.max(&0)).expect("i64 can be converted to usize"))
+        }
+        FieldValue::Uint64(num) => {
+            Some(usize::try_from(*num).expect("i64 can be converted to usize"))
+        }
+        FieldValue::Null => None,
+        FieldValue::Float64(_)
+        | FieldValue::List(_)
+        | FieldValue::Enum(_)
+        | FieldValue::Boolean(_)
+        | FieldValue::String(_) => {
+            panic!("got field value {field_value:#?} in usize_from_field_value which should only ever get Int64, Uint64, or Null")
+        }
+    }
+}
+
 /// If this IRFold has a filter on the folded element count, and that filter imposes
 /// a max size that can be statically determined, return that max size so it can
 /// be used for further optimizations. Otherwise, return None.
@@ -257,11 +278,15 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
                 FoldSpecificFieldKind::Count,
                 Argument::Variable(var_ref),
             ) => {
-                let variable_value = query_arguments[&var_ref.variable_name].as_usize().unwrap();
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
                 Some(variable_value)
             }
             Operation::LessThan(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
-                let variable_value = query_arguments[&var_ref.variable_name].as_usize().unwrap();
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
                 // saturating_sub() here is a safeguard against underflow: in principle,
                 // we shouldn't see a comparison for "< 0", but if we do regardless, we'd prefer to
                 // saturate to 0 rather than wrapping around. This check is an optimization and
@@ -271,7 +296,13 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
             }
             Operation::OneOf(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
                 match &query_arguments[&var_ref.variable_name] {
-                    FieldValue::List(v) => v.iter().map(|x| x.as_usize().unwrap()).max(),
+                    FieldValue::List(v) => v
+                        .iter()
+                        .map(|x| {
+                            usize_from_field_value(x)
+                                .expect("for field value to be coercible to usize")
+                        })
+                        .max(),
                     _ => unreachable!(),
                 }
             }
@@ -341,8 +372,9 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
         // and as an optimization we'd like to stop pulling elements as soon as possible.
         // If we are able to pull more than `max_fold_count_limit + 1` elements,
         // we know that this fold is going to get filtered out, so we might as well
-        // stop materializing its elements early.
-        let mut fold_elements = Vec::with_capacity(*max_fold_count_limit);
+        // stop materializing its elements early. Limit the max allocation size since
+        // it might not always be fully used.
+        let mut fold_elements = Vec::with_capacity((*max_fold_count_limit).min(16));
 
         let mut stopped_early = false;
         for _ in 0..*max_fold_count_limit {
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -513,9 +513,10 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
     let has_tag_on_fold_count = parent_component
         .vertices
         .values()
-        .flat_map(|x| {
-            x.filters.iter().filter(|x| {
-                let Some(Argument::Tag(FieldRef::FoldSpecificField(tagged_fold_count))) = x.right()
+        .flat_map(|vertex| {
+            vertex.filters.iter().filter(|filter| {
+                let Some(Argument::Tag(FieldRef::FoldSpecificField(tagged_fold_count))) =
+                    filter.right()
                 else {
                     return false;
                 };
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -512,9 +512,9 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
         fold.fold_specific_outputs.values().any(|x| *x == FoldSpecificFieldKind::Count);
     let has_tag_on_fold_count = parent_component
         .vertices
-        .iter()
+        .values()
         .flat_map(|x| {
-            x.1.filters.iter().filter(|x| {
+            x.filters.iter().filter(|x| {
                 let Some(Argument::Tag(FieldRef::FoldSpecificField(tagged_fold_count))) = x.right()
                 else {
                     return false;
--- a/trustfall_core/src/interpreter/execution.rs
+++ b/trustfall_core/src/interpreter/execution.rs
@@ -271,8 +271,6 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
     let query_arguments = &carrier.query.as_ref().expect("query was not returned").arguments;
     for post_fold_filter in fold.post_filters.iter() {
         let next_limit = match post_fold_filter {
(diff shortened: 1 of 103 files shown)


trustfall_core/src/interpreter/execution.rs
```diff
@@ -411,6 +474,10 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
     let fold_component = fold.component.clone();
     let fold_eid = fold.eid;
     let max_fold_size = get_max_fold_count_limit(carrier, fold.as_ref());
+    let min_fold_size = get_min_fold_count_limit(carrier, fold.as_ref());
+    let no_outputs_in_fold = fold.component.outputs.is_empty();
+    let has_output_on_fold_count =
+        fold.fold_specific_outputs.values().any(|x| *x == FoldSpecificFieldKind::Count);
```
u9g, line 480: this could be an `is_empty` too but it's not because in the future if there are more FoldSpecificFieldKind's then it would be a problem here, so let's just avoid that right now

trustfall_core/src/interpreter/execution.rs
```diff
@@ -434,7 +501,13 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
         let fold_elements = if fold_exists {
             // N.B.: Note the `?` at the end here!
             //       This lets us early-discard folds that failed a post-processing filter.
-            Some(collect_fold_elements(computed_iterator, &max_fold_size)?)
+            Some(collect_fold_elements(
+                computed_iterator,
+                &max_fold_size,
+                &min_fold_size,
```
u9g, line 507: Why are we passing `Option<usize>` by reference? I left it like this because `max_fold_size` already did this but it's not necessary.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -286,10 +288,52 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
+                Argument::Variable(var_ref),
+            ) => {
+                let variable_value = query_arguments[&var_ref.variable_name].as_usize().unwrap();
+                Some(variable_value)
+            }
+            Operation::GreaterThan(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
+                let variable_value = query_arguments[&var_ref.variable_name].as_usize().unwrap();
+                Some(variable_value + 1)
+            }
+            _ => None,
```
u9g, line 313: We can add a branch like this:
```rust
Operation::NotOneOf(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
  match &query_arguments[&var_ref.variable_name] {
    FieldValue::List(v) => v.iter().map(|x| x.as_usize().unwrap()).max().map(|x| x + 1),
    _ => unreachable!(),
  }
}
```

but let's add more tests first

u9g (commented):
some thoughts

trustfall_core/src/interpreter/execution.rs
```diff
@@ -286,10 +288,52 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
     mut iterator: ContextIterator<'query, Vertex>,
     max_fold_count_limit: &Option<usize>,
+    min_fold_count_limit: &Option<usize>,
+    no_outputs_in_fold: bool,
+    has_output_on_fold_count: bool,
 ) -> Option<Vec<DataContext<Vertex>>> {
+    // We do not apply the min_fold_count_limit optimization if we have an upper bound,
+    // in the form of max_fold_count_limit, because if we have a required upperbound,
+    // then we can't stop at the lower bound, if we are required to observe whether we hit the
+    // upperbound, it's no longer safe to stop at the lower bound.
```
u9g, line 336: is it too much to repeat the same exact thing another time at the end here after the comma?

trustfall_core/src/interpreter/execution.rs
```diff
@@ -286,10 +288,52 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
+        }
+    }
+
+    result
+}
+
 fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
     mut iterator: ContextIterator<'query, Vertex>,
     max_fold_count_limit: &Option<usize>,
+    min_fold_count_limit: &Option<usize>,
```
u9g, line 329: if we make this an isize instead of a usize, then we can make -1 the None case and save half the space, going from 16 for the Option to 8 for an isize, but it might be slightly more confusing.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: Yeah this is a bit hard to read. Can you try rephrasing it a bit, perhaps splitting into two or more sentences with clearer structure?

trustfall_core/src/interpreter/execution.rs
```diff
@@ -318,9 +362,28 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
 
         Some(fold_elements)
     } else {
-        // We weren't able to find any early-termination condition for materializing the fold,
-        // so materialize the whole thing and return it.
-        Some(iterator.collect())
+        let collected = match min_fold_count_limit {
+            // If we have a min_fold_count_limit then it is safe to .take(min_fold_count_limit)
+            // because .take will take at max min_fold_count_limit elements, and if the iterator has
+            // less it will just take all of them.
```
obi1kenobi, line 368:
```suggestion
            // If we have a min_fold_count_limit then it is safe to .take(min_fold_count_limit)
            // because it will take at most min_fold_count_limit elements, and if the iterator has
            // fewer items then it will just take all of them.
```

trustfall_core/src/interpreter/execution.rs
```diff
@@ -318,9 +362,28 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
...
+
+            // This optimization requires that the user can never observe that we didn't fully
+            // iterate the entire iterator. This is only possible if the fold has no outputs and
+            // the user does not have output the count of elements in the fold.
+
```
obi1kenobi, line 373:
```suggestion
            //
            // This optimization requires that the user can never observe that we didn't fully
            // compute the entire `@fold`. This is only possible if the fold has no outputs and
            // the user does not have output the count of elements in the fold.
            //
```
If you're leaving a block comment, always make sure it's linked up even across the blank lines in it.

Also, just a slight wording nit to use user-facing language. "Iterate the iterator" is an implementation detail not visible to the user, they don't see or know about iterators anywhere except as what the `execute_query()` function returns.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -318,9 +362,28 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
...
+            // Additionally, we only apply this optimization if we don't have an upper bound,
+            // in the form of max_fold_count_limit, because if we dont have a required upperbound
+            // can we stop at the lower bound, if we are required to observe whether we hit the
+            // upperbound, it's no longer safe to stop at the lower bound.
```
obi1kenobi, line 377: This also needs some editing for clarity, similar to before.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: I think it's slightly smaller and possibly a bit more efficient as a result: `usize` is one word, `&Option<usize>` is also one word, but `Option<usize>` is at least 1 byte bigger and possibly more.

Bit of a micro-optimization. You did the right thing by matching what the current code was doing. If there was no current code it wouldn't have been a mistake to pass it by value instead of by reference either.

obi1kenobi (commented):
Just some nitpicks on wording. Writing clear and concise comments is an art worth practicing.

The code change looks broadly correct and the test case changes look great.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: I think that's not worth it, too much likelihood of introducing weird bugs.

trustfall_core/src/interpreter/execution.rs
u9g: reworded to now be:
```
If we must collect the fold up to our upperbound of `max_fold_count_limit`,
then we won't use our lowerbound of `min_fold_count_limit`, as by definition
the upperbound will be larger than the lowerbound.
```

trustfall_core/src/interpreter/execution.rs
```diff
@@ -443,6 +506,27 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
...
     let max_fold_size = get_max_fold_count_limit(carrier, fold.as_ref());
+    let min_fold_size = get_min_fold_count_limit(carrier, fold.as_ref());
+    let no_outputs_in_fold = fold.component.outputs.is_empty();
+    let has_output_on_fold_count =
+        fold.fold_specific_outputs.values().any(|x| *x == FoldSpecificFieldKind::Count);
+    let has_tag_on_fold_count = parent_component
+        .vertices
+        .iter()
+        .flat_map(|x| {
+            x.1.filters.iter().filter(|x| {
```
obi1kenobi, line 517: `.iter()` then `x.1` is the same as `.values()` which is more idiomatic Rust

obi1kenobi (commented):
Just some nitpicks to make the Rust more idiomatic, otherwise this is good to go.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: I'm getting tripped up on the "by definition" bit. What definition? Don't we get the upper and lower bounds from separate filters, and it's technically possible that we have a nonsensical query with self-contradicting filters that end up violating this "definition"?

u9g: Shouldn’t that be an impossible filter, and since we solve for the solution space of filters, it would be impossible?

obi1kenobi: Being impossible means it produces no results, not that it can't be executed by a user. The code still has to handle it.

u9g: We can assert on it anyways, if that would be better

obi1kenobi: I don't think we can assert on it — that's what I'm saying. A query with self-contradicting filters is _valid_. It's okay to execute it, and Trustfall must not crash on it. We may one day choose to trigger a lint on such queries, but even then, the query _is allowed_ to be executed, and must still produce the correct — empty — result.

Come to think of it, it's probably best to make some query test cases like that. How about these:
```
{
    Number(min: 30, max: 30) {
        ... on Composite {
            value @output
            
            primeFactor @fold @transform(op: "count") @filter(op: ">=", value: ["$min"]) @filter(op: "<=", value: ["$max"])
        }
    }
}

args:
{
    "min": 2,
    "max": 3,
}
```
- This one should return `{ value: 30 }` because the fold count is 3, which is >= 2 and <= 3.
- The new fold count optimization shouldn't kick in because we still need to check the other filter.

```
{
    Number(min: 30, max: 30) {
        ... on Composite {
            value @output
            
            primeFactor @fold @transform(op: "count") @filter(op: ">=", value: ["$min"]) @filter(op: "<=", value: ["$max"])
        }
    }
}

args:
{
    "min": 3,
    "max": 2,
}
```
- This one should return no results because the filter conditions are `>= 3 && <= 2`.
- The new fold count optimization probably shouldn't kick in here either, for the same reason.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -317,10 +319,56 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
+                Some(variable_value)
+            }
+            Operation::GreaterThan(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
+                Some(variable_value + 1)
```
obi1kenobi, line 346: micro-fix:
```suggestion
                Some(variable_value.saturating_add(1))
```
so we don't overflow and panic

trustfall_core/src/interpreter/execution.rs
u9g: Added a test for this

trustfall_core/src/interpreter/execution.rs
```diff
@@ -317,10 +317,56 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
+                Some(variable_value.saturating_add(1))
+            }
+            // This optimization is only valid if we know that every
+            // filter applied to the folded element count is a comparison
+            // that we can determine the number of elements needed to satisfy
+            // the filter.
+            _ => return None,
```
u9g, line 348: this actually doesn't fix any tests, it would be good to introduce a failing test that is fixed by this so we know that this `return None` is doing the right thing

trustfall_core/src/interpreter/execution.rs
```diff
@@ -443,6 +502,25 @@ fn compute_fold<'query, AdapterT: Adapter<'query> + 'query>(
...
+            else {
+                return false;
+            };
+
+            tagged_fold_count.fold_root_vid == fold.to_vid
+                && tagged_fold_count.fold_eid == fold.eid
+                && tagged_fold_count.kind == FoldSpecificFieldKind::Count
+        })
+    });
+    let safe_to_skip_part_of_fold =
```
u9g, line 522: is `safe_to_skip_part_of_fold` too non-specific as to which part we can skip?

trustfall_core/src/interpreter/execution.rs
u9g: test case added for this: https://github.com/obi1kenobi/trustfall/pull/431

trustfall_core/src/interpreter/execution.rs
```diff
@@ -317,10 +317,56 @@ fn get_max_fold_count_limit(carrier: &mut QueryCarrier, fold: &IRFold) -> Option
...
+            Operation::GreaterThan(FoldSpecificFieldKind::Count, Argument::Variable(var_ref)) => {
+                let variable_value =
+                    usize_from_field_value(&query_arguments[&var_ref.variable_name])
+                        .expect("for field value to be coercible to usize");
+                Some(variable_value.saturating_add(1))
+            }
+            // This optimization is only valid if we know that every
+            // filter applied to the folded element count is a comparison
+            // that we can determine the number of elements needed to satisfy
+            // the filter.
```
obi1kenobi, line 347: this portion of this comment doesn't really make sense to me:
```
a comparison that we can determine the number of elements needed to satisfy the filter.
```

To simplify, you can out "this optimization is only valid if" and replace it with "every filter on the fold count must ..."

But also, I'm not sure this comment is at the right level of abstraction. This comment makes a general statement about "this optimization" where the optimization is never defined — the function's docs say "further optimizations." It doesn't say anything about this match branch, so it would be hard for the reader to connect all the clues to figure out what's going on.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: nicely done!

trustfall_core/src/interpreter/execution.rs
```diff
@@ -350,9 +396,22 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
...
+        let collected = match min_fold_count_limit {
+            // Queries that do not observe the fold count nor any fold contents may be able to
+            // be optimized by only partially expanding the fold, just enough to check any filters
+            // that may be applied to the fold count.
+            //
+            // For example, if `@filter(op: ">", value: ["$ten"])` is our only filter on the count
+            // of the fold, we can stop computing the rest of the fold after seeing we have 11 elements.
+            Some(min_fold_count_limit) if safe_to_skip_part_of_fold => {
+                iterator.take(*min_fold_count_limit).collect()
+            }
```
obi1kenobi, line 408: Do you think it's worth tweaking the implementation here a bit to simplify the logic?

Here's what I noticed: `safe_to_skip_part_of_fold = false` with `min_fold_count_limit = Some(...)` is the same as `safe_to_skip_part_of_fold = true` with `min_fold_count_limit = None`. Also, `safe_to_skip_part_of_fold` has a pretty awkward name, which is usually a sign that the concept it represents is muddy as well.

In that light, it seems weird that we're splitting the logic so that the caller of this function prepares the two values, but isn't allowed to combine them. Perhaps we can eliminate `safe_to_skip_part_of_fold` completely by making the caller turn the `min_fold_count_limit` to `None` if the fold expansion couldn't be short-circuited?

In that case, the bulk of this comment would be best moved outside into the caller as well, since that's where most of the logic is on _whether_ this optimization can be applied. This function just ends up doing what it's told.

trustfall_core/src/interpreter/execution.rs
obi1kenobi: The name isn't great. But see the comment above on an idea how we can skip it entirely.

Meta: try naming things for what they _represent_ and not for _what should be done with them_. Right now, you tend to give a lot of imperative names like that, but those tend to age poorly because they might end up being used differently than originally intended. Names that describe what the value _represents_, not _how it's going to be used_, will age much better. Try thinking of some names like that and maybe put a few ideas in reply if you want feedback?

trustfall_core/src/interpreter/execution.rs
u9g: Hm, you're right. Maybe something like `If we find a filter that we can't evaluate partially evaluate, fail early.` Would be a better comment.

obi1kenobi: Much better! Small nit: "partial evaluation" is also the term for an unrelated idea, which might confuse the reader. Perhaps "filter that would require evaluating the fold past its minimum count" or something?

I don't love my suggestion, perhaps you can come up with something even better.

u9g: I see what you're saying, `If we find a filter that requires us to evaluate the entire fold, fail early.` is better I think since it doesn't require knowing what we are partially evaluating. Although this gives the mistaken impression that every fold count filter that gets here will need to be fully evaluated which isn't necessarily the case. Another try would be `If we don't know how many elements of the fold are required to satisfy this filter, fail early.`

obi1kenobi: The latter is good, I like that. "Requires us to evaluate the entire fold" is technically incorrect as you point out.

trustfall_core/src/interpreter/execution.rs
u9g: Right, I removed this `if safe_to_skip_part_of_fold` from `collect_fold()` and put it in the caller, and combined the boolean checks, this way `collect_fold()` only has to execute on the option, it doesn't have to deal with extra conditions if the option already has one built in.

trustfall_core/src/interpreter/execution.rs
u9g: Right, to name this variable something better, I think `fold_values_not_observed` or `fold_values_ignored` would suffice since that's what this boolean represents is that the fold values are not observed, neither the final count of values not the individual values by `@fold`'ing on them.

trustfall_core/src/interpreter/execution.rs
u9g: Okay great, updated to that

obi1kenobi (approved):

PR #425, opened 2023-08-02
Add negative filter tests

u9g:


trustfall_core/src/interpreter/execution.rs
```diff
@@ -241,6 +241,24 @@ fn construct_outputs<'query, AdapterT: Adapter<'query>>(
...
 }
 
+/// takes a FieldValue and produces a usize clamped to 0 and usize::MAX
+fn usize_from_field_value(field_value: &FieldValue) -> Option<usize> {
+    match field_value {
+        FieldValue::Int64(num) => {
+            Some(usize::try_from(*num.max(&0)).expect("i64 can be converted to usize"))
+        }
+        FieldValue::Uint64(num) => {
+            Some(usize::try_from(*num).expect("i64 can be converted to usize"))
```
u9g, line 251: usize should be bigger than both of these right? there's no need to also .min this with `usize::MAX` I think

obi1kenobi: It's not actually guaranteed to, it's possible that `usize` is 32-bit e.g. on `wasm32` platforms whereas `u64` is always 64-bit. Your call on whether we sort this out now or not.

obi1kenobi: But either update the docstring or the code, because right now the docstring says this is clamped to `usize::MAX` when the code doesn't actually clamp it.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -241,6 +241,24 @@ fn construct_outputs<'query, AdapterT: Adapter<'query>>(
     }))
 }
 
+/// takes a FieldValue and produces a usize clamped to 0 and usize::MAX
```
obi1kenobi, line 244:
```suggestion
/// Takes a FieldValue and produces a usize clamped between 0 and `usize::MAX`.
```
Docstrings should always be capitalized and use code blocks for code.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -241,6 +241,24 @@ fn construct_outputs<'query, AdapterT: Adapter<'query>>(
...
+        }
+        FieldValue::Uint64(num) => {
+            Some(usize::try_from(*num).expect("i64 can be converted to usize"))
+        }
+        FieldValue::Float64(_)
+        | FieldValue::List(_)
+        | FieldValue::Enum(_)
+        | FieldValue::Boolean(_)
+        | FieldValue::String(_)
+        | FieldValue::Null => None,
```
obi1kenobi, line 258: For `null` it's fine to return `None` (imagine a tag being used), but for everything else we should panic instead because that indicates a bug in the type-checker.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -298,7 +326,7 @@ fn collect_fold_elements<'query, Vertex: Clone + Debug + 'query>(
         // If we are able to pull more than `max_fold_count_limit + 1` elements,
         // we know that this fold is going to get filtered out, so we might as well
         // stop materializing its elements early.
-        let mut fold_elements = Vec::with_capacity(*max_fold_count_limit);
+        let mut fold_elements = Vec::with_capacity((*max_fold_count_limit).min(16));
```
obi1kenobi, line 329: please update the comment to explain why `.min(16)`, it isn't obvious.

trustfall_core/test_data/tests/valid_queries/fold_count_filter_eq.graphql.ron
obi1kenobi, line 1: Please name the file something that more accurately communicates what's being tested, right now it isn't obvious that the point of the test is to use a negative number as the argument.

trustfall_core/test_data/tests/valid_queries/fold_count_filter_eq.graphql.ron
```diff
@@ -0,0 +1,19 @@
+TestGraphQLQuery (
+    // We use a negative number as a filter to ensure that we properly handle the negative number case in get_max_fold_count_limit.
```
obi1kenobi, line 2: A few issues here:
- This line is too long. We don't currently have a linter set up, but please try to keep lines below 100 chars in all files to make sure they are readable.
- The comment references implementation details that may change and that ultimately aren't important. The test ensures that comparisons with negative numbers on count work _in general_, and would catch any failures even if they aren't in `get_max_fold_count_limit()`. I wouldn't name that function or any other implementation details here.
- `get_max_fold_count_limit()` is code and a function. Please try to always put code in backticks and have function names followed by `()` to show they are functions.

u9g: I spruced up the comments in the tests to look like this:

`// Ensure that we properly handle negative numbers in a `not_one_of` filter on a folded edge's count.`

I believe this is the longest one, and it's 101 characters including `// ` so I think that's fine for now. I removed any mentions to internals in tests, good call.

trustfall_core/src/interpreter/execution.rs
u9g: I update3d the docstring for now to just not say it's clamped to usize::MAX, instead it just says that we return a positive usize.

trustfall_core/src/interpreter/execution.rs
```diff
@@ -241,6 +241,27 @@ fn construct_outputs<'query, AdapterT: Adapter<'query>>(
     }))
 }
 
+/// Takes a FieldValue and returns a positive usize. If the FieldValue is a Null,
+/// returns None.
```
obi1kenobi, line 245:
```suggestion
/// Extracts numeric `FieldValue` into a `usize`, clamping negative numbers to 0.
/// Returns `None` on `FieldValue::Null`, and panics otherwise.
```

obi1kenobi (approved):

PR #429, opened 2023-08-04
Add impossible filter and dominated filter tests

u9g:


trustfall_core/test_data/tests/valid_queries/fold_count_filter_with_overrode_filter.graphql.ron
```diff
@@ -0,0 +1,22 @@
+TestGraphQLQuery (
+    // Ensure that we properly handle two of the same type of filters
+    // but one of the filters beat's the other. In this case filtering
+    // by greater than two is beat by filtering for greater than three.
```
obi1kenobi, line 4:
```suggestion
    // but one of the filters dominates the other. In this case filtering
    // by greater than two is dominated by filtering for greater than three.
```

this is the technical term you were looking for here

also the apostrophe wasn't needed in the original

trustfall_core/test_data/tests/valid_queries/fold_count_filter_with_overrode_filter.graphql.ron
```diff
@@ -0,0 +1,22 @@
...
+    schema_name: "numbers",
```
obi1kenobi, line 5:
```suggestion
    // 
    // This test guards against an accidental min/max mixup in the implementation
    // of the optimization that allows us to stop expanding folds early.
    // We should only stop when the dominating filter is satisfied,
    // not when the dominated filter is satisfied. If such a bug were to happen,
    // this query would return `{ "value": 30 }` instead of having no results.
    schema_name: "numbers",
```

obi1kenobi: A bit of additional explanation to make sure we remember why we added this test case.

trustfall_core/test_data/tests/valid_queries/fold_count_filter_with_overrode_filter.graphql.ron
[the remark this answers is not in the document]
u9g: Done!

obi1kenobi (approved):

PR #431, opened 2023-08-05
Add a test for ignoring necessary filters

u9g:
this adds a test for https://github.com/obi1kenobi/trustfall/pull/423?notification_referrer_id=NT_kwDOApfigbM3MjU5ODY2MjY3OjQzNTA4MzUz#discussion_r1284964824

### DIFF
--- a/trustfall_core/test_data/tests/valid_queries/fold_count_filter_on_a_tag.graphql.ron
+++ b/trustfall_core/test_data/tests/valid_queries/fold_count_filter_on_a_tag.graphql.ron
@@ -9,30 +9,18 @@ TestGraphQLQuery (
     schema_name: "numbers",
     query: r#"
 {
-    Number(min: 6, max: 6) {
-        ... on Composite {
-
-            value @output
-
-            predecessor { # five
-                predecessor { # four
-                    predecessor { # three
-                        predecessor { # two
-                            predecessor { # one
-                                value @tag(name: "one")
-                            }
-                        }
-                    }
-                }
-            }
-
+    Two {
+        value @tag(name: "two")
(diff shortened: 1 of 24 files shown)


trustfall_core/test_data/tests/valid_queries/fold_count_filter_on_a_tag.graphql.ron
```diff
@@ -0,0 +1,38 @@
...
+                    }
+                }
+            }
+
+            primeFactor @fold @transform(op: "count")
+                        @filter(op: "<", value: ["%one"])
+                        @filter(op: ">=", value: ["$zero"])
+        }
+    }
+}"#,
```
obi1kenobi, line 34: This query is quite long, which makes its test files large (the trace alone is 700+ lines) and also would make debugging unnecessarily complex.

Here's a simpler variant to consider. I've also swapped the order of the filters so that the passing one gets evaluated first (because right now we run them in the same order as in the query):
```suggestion
{
    Two {
        value @tag(name: "two")

        multiple(max: 3) {
            value @filter(op: "=", value: ["$six"]) @output

            primeFactor @fold @transform(op: "count")
                        @filter(op: ">=", value: ["$zero"])
                        @filter(op: "<", value: ["%two"])
        }
    }
}"#,
```

What do you think?

u9g: You're right, this is a good change. I didn't even think to check the trace file size, but it's now down by 300 lines to only 400 after your change

obi1kenobi (approved):

PR #434, opened 2023-08-06
Add external types exposed through public api check for CI

u9g:


u9g:
Setup closely mirrors https://github.com/awslabs/cargo-check-external-types/blame/main/.github/workflows/ci.yml#L110

.github/workflows/ci.yml
```diff
@@ -172,6 +176,28 @@ jobs:
...
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: "nightly-2023-05-31"
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Check Exported API
+        run: |
+          cargo install --locked cargo-check-external-types
```
obi1kenobi, line 197: This is going to rebuild on every PR update, which is going to be horrifically slow. The tool's doesn't offer prebuilt binaries, and it's not part of `cargo-quickinstall`'s prebuilt list either.

Almost tempted to not use it over this, to be honest.

What do you think?

u9g: I'm not super upset over it as it takes approximately the same 2 minutes to run the tests. We can also either ask them to offer prebuilt binaries or just build it ourselves on every commit in a different repo similar to what cargo-quickinstall does.

Is there any way we can ask them to add it to cargo-quickinstall?

obi1kenobi:
Not sure, check that project's repo? There might be a way to request new
crates to be added.

PR #435, opened 2023-08-06
Move to our own Type type for everything that's exposed in trustfall_core

u9g:
No longer re-export a 3rd party crate's `Type` type for Trustfall output or variable types. Instead of that, use our own.


trustfall_core/src/frontend/mod.rs
```diff
@@ -1161,7 +1160,7 @@ where
                 let tag_field = ContextField {
                     vertex_id: current_vid,
                     field_name: subfield.name.clone(),
-                    field_type: subfield_raw_type.to_owned(),
+                    field_type: subfield_raw_type.clone(),
```
u9g, line 1163: I changed this to be the same as the `.clone()`'s above, but not sure if it's necessary

trustfall_core/src/ir/indexed.rs
```diff
@@ -29,8 +29,6 @@ pub struct IndexedQuery {
 pub struct Output {
     pub name: Arc<str>,
 
-    #[serde(serialize_with = "crate::ir::serialization::serde_type_serializer")]
```
u9g, line 32: We don't necessarily have to get rid of these, but I don't think it's good to keep them if we already have the serialize and deserialize directly on `Type`

trustfall_core/src/ir/indexed.rs
```diff
@@ -101,13 +99,10 @@ fn get_output_type(
 ) -> Type {
     let mut wrapped_output_type = field_type.clone();
     if component_optional_vertices.contains(&output_at) {
-        wrapped_output_type.nullable = true;
+        wrapped_output_type = wrapped_output_type.with_nullability(true);
```
u9g, line 102: the way it stands now `Type` is immutable. I think this is one of a few cases in the entire codebase that we would have benefited from it not being immutable, although it was an easy change, and immutability is good, I think.

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// use trustfall_core::ir::ty::{*};
+    ///
+    /// let individual_ty = Type::new("Int!").unwrap();
+    /// assert!(matches!(individual_ty.value(), InnerType::NameOfType("Int")));
+    ///
+    /// let list_ty = Type::new("[Int!]!").unwrap();
+    /// let inner_ty_for_assert = Type::new("Int!").unwrap();
+    /// assert!(matches!(list_ty.value(), InnerType::ListInnerType(inner_ty) if inner_ty == inner_ty_for_assert));
+    /// ```
+    pub fn value(&self) -> InnerType<'_> {
```
u9g, line 121: Is `value` a bad name for this? Another candidate for this could be `inner_value`. Or, we could separate out what this enum gives us to two separate methods but I think it's much more useful and idiomatic this way, despite us doing more work in some cases to clone the List's inner type for the List() even if that branch is ignored. (although maybe that's optimized away? not sure)

trustfall_core/src/ir/types.rs
```diff
@@ -135,25 +127,24 @@ pub(crate) fn get_base_named_type(ty: &Type) -> &str {
...
-            is_scalar_only_subtype(parent_type, maybe_subtype)
-        }
-        (BaseType::Named(..), BaseType::List(..)) | (BaseType::List(..), BaseType::Named(..)) => {
-            false
+    match (parent_type.value(), maybe_subtype.value()) {
+        (InnerType::NameOfType(parent), InnerType::NameOfType(subtype)) => parent == subtype,
+        (InnerType::ListInnerType(parent_type), InnerType::ListInnerType(maybe_subtype)) => {
+            is_scalar_only_subtype(&parent_type, &maybe_subtype)
         }
+        (InnerType::NameOfType(_), InnerType::ListInnerType(_))
```
u9g, line 139: Should I just use `_ => false` here instead?

trustfall_core/src/ir/types.rs
```diff
@@ -166,27 +157,29 @@ pub(crate) fn is_scalar_only_subtype(parent_type: &Type, maybe_subtype: &Type) -
...
                 None
             }
         }
-        (BaseType::List(left), BaseType::List(right)) => intersect_types(left, right)
-            .map(|inner| Type { base: BaseType::List(Box::new(inner)), nullable }),
-        (BaseType::Named(_), BaseType::List(_)) | (BaseType::List(_), BaseType::Named(_)) => None,
+        (InnerType::ListInnerType(left), InnerType::ListInnerType(right)) => {
+            intersect_types(left, right).map(|inner| Type::new_list_type(inner, nullable))
+        }
+        (InnerType::NameOfType(_), InnerType::ListInnerType(_))
```
u9g, line 173: Same as above, should I just use `_ => None` here?

trustfall_core/src/schema/mod.rs
```diff
@@ -397,11 +402,11 @@ fn check_type_and_property_and_edge_invariants(
 
                     // Check that the edge field doesn't have
                     // a list-of-list or more nested list type.
-                    match &field_type.base {
-                        BaseType::Named(_) => {}
-                        BaseType::List(inner) => match &inner.base {
-                            BaseType::Named(_) => {}
-                            BaseType::List(_) => {
+                    match &field_type.value() {
```
u9g, line 405: this can also be expressed as
```rs
if matches!(&field_type.value(), InnerType::ListInnerType(ty) if matches!(ty.value(), InnerType::ListInnerType(_))) {
```

but I don't think it's better, and it's out of scope

trustfall_core/src/frontend/mod.rs
```diff
@@ -1114,7 +1113,7 @@ where
                         .or_default()
                         .push(subfield_name.clone());
 
-                    (subfield_name, subfield_raw_type, SmallVec::from([subfield]))
+                    (subfield_name, subfield_raw_type.clone(), SmallVec::from([subfield]))
```
u9g, line 1116: We can get a much cheaper clone if we place this in an Rc, infact, we can do this everywhere if we keep `Type` immutable.

trustfall_core/src/ir/ty.rs
obi1kenobi: Yeah, it's not a great name.

I'm also really not sure about `InnerType` as a whole. When you find it difficult to come up with a good name for a concept, oftentimes it's an indicator of a confusing or muddled design.

obi1kenobi (commented):
I struggled with this review, because _something_ doesn't feel right about the new design here, but I can't put my finger on what. I think I'm getting tripped up on the naming, and failing to see the forest for the trees...

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,203 @@
...
+
+impl<'a> PartialEq<&'a Type> for Type {
+    fn eq(&self, other: &&'a Type) -> bool {
+        self.ty == other.ty
+    }
+}
+
+pub(crate) fn from_type(ty: &GQLType) -> Type {
+    Type { ty: ty.clone() }
+}
```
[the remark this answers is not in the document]
u9g: I wanted to but I can’t pub(crate) it because the trait is public (I think)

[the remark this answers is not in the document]
obi1kenobi: Probably best as an associated function on `Type` then, not a free function.

trustfall_core/src/ir/ty.rs
u9g: Did you have a better interface in mind for this? I think that this is the most ergonomic way to expose the data that the user wants to know of is this a NameOfType (Named Type) or a ListInnerType (List). For the most part this is what we already have with the BaseType enum, so it doesn’t require too much code changes, but if you have a better design you can see, please share.

obi1kenobi: It would require a lot of testing, but if you're up for it, here's a fairly compact and efficient `Type` implementation with no pointer chasing in the common path, and with all the internals hidden:
```rust
pub struct Type {
    base: Arc<str>,
    modifiers: Modifiers,
}

// This type is meant to be tiny: discriminant + 1 pointer/u64 union
enum Modifiers {
    Inline(InlineModifiers),
    Extended(Box<ExtendedModifiers>),
}

struct InlineModifiers {
    mask: u64,  // space for ~30 levels of list nesting
}

impl InlineModifiers {
    const NON_NULLABLE_MASK: u64 = 1;
    const LIST_MASK: u64 = 2;

    fn is_nullable(&self) -> bool {
        (self.mask & Self::NON_NULLABLE_MASK) == 0
    }

    fn is_list(&self) -> bool {
        (self.mask & Self::LIST_MASK) != 0
    }

    fn as_list(&self) -> Option<InlineModifiers> {
        self.is_list().then_some(|| {
            InlineModifiers::new(self.mask >> 2)
        })
    }
}

// In case someone really wants more than ~30 levels of nested lists in their type...
struct ExtendedModifiers {
    // ... TODO ...
}

impl Type {
    pub fn nullable(&self) -> bool {
        self.modifiers.is_nullable()
    }

    pub fn is_list(&self) -> bool {
        self.modifiers.is_list()
    }

    pub fn list_of(&self) -> Option<Self> {
        self.modifiers.as_list().map(|modifiers| {
            Self {
                base: self.base.clone(),
                modifiers,
            }
        })
    }

    pub fn base_type(&self) -> &str {  // this ensures the Arc<str> is an implementation detail
        self.base.as_ref()
    }
}
```

nvzqz: @obi1kenobi I think that initial definition of `Type` is a good start. If we want to optimize even further, there's some options to reduce `Type` from 4 words. These are rather tricky to get right, and we can optimize piecemeal, so **I recommend against doing any of these in this commit**.

To preface, we have an abundance of free real estate for tags:
- The most significant bit of `str` length because it cannot exceed `isize::MAX`
- The bottom bits of pointers because we can enforce the alignment. For example:
    - 2 byte align: 1 bit
    - 4 byte align: 2 bits
    - 8 byte align: 3 bits

Given the space for tags, we have various options for optimizing `Type`:
- Make `Type` 3 words by making `Modifiers` a tagged pointer, leaving us 63 bits for `InlineModifiers`
- Make `Type` 2 words by storing `ExtendedModifiers` as the header for the base name `str`
- Make `Type` 1 word by moving `str` length into the allocation along with the optional `ExtendedModifiers` header
- Compile-time types!
    - Tag `Type` to indicate `&'static` instead of `Arc`
    - Allows commonly-known types to have free `Clone` and no heap alloc

obi1kenobi: Great to know we'll be set up for future optimizations, should they prove necessary.

Let's start with the initial `Type` layout (which should already eliminate a lot of pointer-chasing and be faster for it), and we can optimize further in the future — which won't even be a breaking change from here onward!

@u9g do you feel like you can take this initial sketch of a design for `Type` and run with it from here?

nvzqz: Bonus optimization: `Type` could also store tiny strings inline, up to 7 bytes for 1 word `Type`

u9g: Yes, I'll start working with the initial design and we can optimize it later :)

trustfall_core/src/ir/indexed.rs
u9g: It's a breaking change for any users, but we now implement this directly on our `Type` struct so It's probably fine.

trustfall_core/src/ir/ty.rs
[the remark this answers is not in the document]
u9g: Wouldn't they be side-by-side in the `types` directory too though?

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,203 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum InnerType<'a> {
+    NameOfType(&'a str),
+    ListInnerType(Type),
+}
+
+/// A backing-storage independent immutable representation of a GraphQL type.
+impl Type {
+    /// Creates a new [`Type`] from a string.
+    /// Returns `None` if the string is not a valid GraphQL type.
```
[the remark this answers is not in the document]
u9g: Yes, if we want to write our own graphql type parser. It wouldn't be too hard I just didn't want to do it in this pr. `async-graphql-parser`'s `Type::new()` returns `Option<Type>` so we can't get better error messages yet.

trustfall_core/src/ir/ty.rs
u9g: The initial implementation has landed!

trustfall_core/src/ir/ty.rs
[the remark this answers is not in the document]
obi1kenobi: At that point we could come up with some better names for the files :)

trustfall_core/src/ir/mod.rs
```diff
@@ -232,8 +220,7 @@ pub enum FoldSpecificFieldKind {
     Count, // Represents the number of elements in an IRFold's component.
 }
 
-static NON_NULL_INT_TYPE: Lazy<Type> =
-    Lazy::new(|| Type { base: BaseType::Named(Name::new("Int")), nullable: false });
+static NON_NULL_INT_TYPE: Lazy<Type> = Lazy::new(|| Type::new_named_type("Int", false));
```
obi1kenobi, line 223: this and other common types like `String!` for `__typename` are probably best moved into the new `types` module I proposed

trustfall_core/src/ir/mod.rs
```diff
@@ -636,7 +623,7 @@ impl<LeftT: NamedTypedValue> Operation<LeftT, Argument> {
                 // For the operands relative to each other, nullability doesn't matter,
                 // but the rest of the type must be the same.
                 let right_type = right_type.unwrap();
-                if are_base_types_equal_ignoring_nullability(&left_type.base, &right_type.base) {
+                if are_base_types_equal_ignoring_nullability(left_type, right_type) {
```
obi1kenobi, line 626: This function and similar ones below might be good to refactor and possibly either expose as methods directly, or reimplement using appropriate helper methods.

They were free functions because we couldn't implement methods on a foreign `Type` type. Now the type is ours so we can do better.

trustfall_core/src/ir/mod.rs
```diff
@@ -712,22 +699,21 @@ impl<LeftT: NamedTypedValue> Operation<LeftT, Argument> {
...
-                        )])
-                    }
+                let inner_type = if let Some(list) = left_type.as_list() {
+                    Ok(list)
+                } else {
+                    Err(vec![FilterTypeError::ListFilterOperationOnNonListField(
+                        self.operation_name().to_string(),
+                        left.named().to_string(),
+                        left_type.to_string(),
+                    )])
```
obi1kenobi, line 709: This seems like a manual implementation of `Option::ok_or_else()`

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+    fn is_list(&self) -> bool {
+        (self.mask & Self::LIST_MASK) != 0
+    }
+
+    fn as_list(&self) -> Option<Modifiers> {
+        self.is_list().then_some(Modifiers { mask: self.mask >> 2 })
+    }
+
+    fn at_max_list_depth(&self) -> bool {
+        (self.mask & (Self::MAX_LIST_DEPTH_MASK)) == (Self::MAX_LIST_DEPTH_MASK)
```
obi1kenobi, line 39: some of the parens are a bit excessive :)
```suggestion
        (self.mask & Self::MAX_LIST_DEPTH_MASK) == Self::MAX_LIST_DEPTH_MASK
```

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+    ///
+    /// let outer_nullable = true;
+    /// let ty = Type::new_list_type(inner_ty, outer_nullable);
+    ///
+    /// assert_eq!(ty.to_string(), "[String!]");
+    /// assert_eq!(ty, Type::new("[String!]").unwrap());
+    /// ```
+    pub fn new_list_type(inner_type: Self, nullable: bool) -> Self {
+        if inner_type.modifiers.at_max_list_depth() {
+            panic!("too many nested lists");
```
obi1kenobi, line 96: Panics with a constant error message are generally best avoided because they make debugging hard.

Try to include relevant information about the exact inputs that caused this.

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+            if !ty_inside_list.nullable {
+                mask |= Modifiers::NON_NULLABLE_MASK << i;
+            }
+            base = &ty_inside_list.base;
+        }
+
+        let BaseType::Named(name) = base else {
+            unreachable!(
+                "should be impossible to get a non-named type after looping through all list types"
+            )
```
obi1kenobi, line 217: in both the panics, include the input that caused the problem and as much relevant debugging information as possible

think of what you'd like to know next if this panic is triggered, then include that data up front to save us a few steps

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+        // left
+        {
+            let mut current = Some(self.modifiers.clone());
+            while let Some(mods) = current {
+                if mods.is_list() {
+                    write!(f, "[")?;
+                }
+                current = mods.as_list();
+            }
+        }
```
obi1kenobi, line 234: Perhaps indicative of a missing helper method for counting the number of nested lists?

Should be just a mask and a popcount.

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        write!(f, "{}", self.base)?;
+
+        let mut builder = String::new();
```
obi1kenobi, line 238: Consider using `String::with_capacity()` here to ensure we don't have to reallocate. The capacity should be related to the pop count of the mask.

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+            fn expecting(&self, formatter: &mut Formatter) -> fmt::Result {
+                formatter.write_str("GraphQL type")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                let ty = Type::new(s)
+                    .ok_or_else(|| serde::de::Error::custom("not a valid GraphQL type"))?;
```
obi1kenobi, line 287: Will serde include the input type that isn't valid? If not, can you include it in the error message explicitly?

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+    #[test]
+    fn max_allowed_nested_lists() {
+        let my_str = format!(
+            "{}String{}",
+            "[".repeat(Modifiers::MAX_LIST_DEPTH as usize),
+            "]".repeat(Modifiers::MAX_LIST_DEPTH as usize)
+        );
+        let constructed_type = Type::new(&my_str).unwrap().modifiers;
+        assert!(constructed_type.at_max_list_depth());
+    }
```
obi1kenobi, line 311: Can you add a test like this but that also uses nullability?

Remember that a list contents' nullable bit is shifted one to the left from the last list bit. So while I think the implementation is probably correct, it's in principle possible that the nullability bit overflows even though the max list bit is fine.

trustfall_core/src/ir/ty.rs
```diff
@@ -0,0 +1,332 @@
...
+                    .ok_or_else(|| serde::de::Error::custom("not a valid GraphQL type"))?;
+                Ok(ty)
+            }
+        }
+
+        deserializer.deserialize_str(TypeDeserializer)
+    }
+}
+
+#[cfg(test)]
```
obi1kenobi, line 296: This needs lots more tests. A few examples off the top of my head:
- Complex types with mixed nullability across multiple nested lists get parsed, represented, and `.to_string()`-ed correctly.
- More in-depth tests for `.to_string()` in general.
- Tests for the methods that allow building up types by wrapping existing types in a list and/or setting nullability.

trustfall_core/src/ir/types.rs
```diff
@@ -95,30 +93,13 @@ impl NamedTypedValue for Argument {
...
-
-pub(crate) fn is_base_type_orderable(operand_type: &BaseType) -> bool {
-    match operand_type {
-        BaseType::Named(name) => {
-            name == "Int" || name == "Float" || name == "String" || name == "DateTime"
-        }
-        BaseType::List(l) => is_base_type_orderable(&l.base),
-    }
+pub(crate) fn are_base_types_equal_ignoring_nullability(left: &Type, right: &Type) -> bool {
+    left.base_named_type() == right.base_named_type()
```
obi1kenobi, line 97: This is not a correct implementation. It does not match the semantics of the original function. Please write tests to catch the bug then update the implementation.

Please also review your other implementation changes in this module for similar errors.

trustfall_core/src/ir/types.rs
```diff
@@ -201,32 +179,31 @@ pub fn is_argument_type_valid(variable_type: &Type, argument_value: &FieldValue)
...
         }
         FieldValue::String(_) => {
             // This is a valid value only if the type is String, ignoring nullability.
-            matches!(&variable_type.base, BaseType::Named(n) if n == "String")
+            !variable_type.is_list() && variable_type.base_named_type() == "String"
         }
         FieldValue::Boolean(_) => {
             // This is a valid value only if the type is Boolean, ignoring nullability.
-            matches!(&variable_type.base, BaseType::Named(n) if n == "Boolean")
+            !variable_type.is_list() && variable_type.base_named_type() == "Boolean"
```
obi1kenobi, line 198: This is also not correct.

obi1kenobi (commented):
It's a step in the right direction, but there's lot of complex code with very insufficient testing, and at least a few serious bugs have snuck in as a result.

trustfall_core/src/ir/ty.rs
u9g: > Can you add a test like this but that also uses nullability?
Yes
> it's in principle possible that the nullability bit overflows even though the max list bit is fine.
I don't believe so. Here's the bit representation for that test `101010101010101010101010101010101010101010101010101010101010`
Here it is if there is an `!` at the very end too
`101010101010101010101010101010101010101010101010101010101011`

As you can see, only the very last bit on the very right changed from a zero to one, however I have added a mask representation test to ensure that doesn't happen.

u9g: Ohh interesting, yeah you're right. If the innermost type is non-nullable, then we get the bitmask of:
`1101010101010101010101010101010101010101010101010101010101010`
(with no non-nulls)
`101010101010101010101010101010101010101010101010101010101010`
this does indeed add a new digit but we fit it fine.

PR #444, opened 2023-08-10
Allow any script to be executed anywhere inside the repo

era:
I noticed most tests were doing `cd "$(git rev-parse --show-toplevel)/trustfall_testbin"`, except those that receive paths as parameters. My guess is those scripts were not doing that because the user may give relative paths and changing the directory will break user assumptions about it.

For those scripts, I transform any path the user set as arguments into an absolute path, so we can safely move directories.

I tested this by passing absolute paths and relative paths on Linux.

### DIFF
--- a/scripts/create_execution_error_data.sh
+++ b/scripts/create_execution_error_data.sh
@@ -6,7 +6,17 @@ set -euo pipefail
 # Get the absolute path of the repo.
 REPO="$(git rev-parse --show-toplevel)"
 
+INPUT_FILES=()
+
 for INPUT_FILE in "$@"; do
+    # makes sure we are always using absolute path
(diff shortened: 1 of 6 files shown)


scripts/create_execution_error_data.sh
```diff
@@ -6,7 +6,17 @@ set -euo pipefail
 # Get the absolute path of the repo.
 REPO="$(git rev-parse --show-toplevel)"
 
+INPUT_FILES=()
+
 for INPUT_FILE in "$@"; do
+    # makes sure we are always using absolute path
+    INPUT_FILES+=("$(cd "$(dirname "$1")"; pwd)/$(basename "$1")")
```
obi1kenobi, line 13: This is a nice trick, I'll have to remember that! 💯

obi1kenobi (approved):

PR #449, opened 2023-08-14
Add an iterator over all the required properties of a Vertex

era:
In order to enable adapters to only fetch the required properties (useful for network calls, such as GraphQL adapters) we need to expose all the required properties of that vertex. 

This PR introduces a method to retrieve the information (output, filtered, and tagged properties). It also introduces a new struct to represent those properties.

It also introduces tests for the new feature.


trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -14,6 +14,17 @@ use crate::{
 
 use super::{dynamic::DynamicallyResolvedValue, CandidateValue, EdgeInfo, Range};
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RequiredProperty {
```
era, line 18: I need to re-home this, it's just on the same file as the `VertexInfo` for convenience when I was coding it

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -14,6 +14,17 @@ use crate::{
...
+    pub name: Arc<str>,
```
era, line 19: Not sure if we should add more fields such as property type or not

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -23,6 +34,8 @@ pub trait VertexInfo: super::sealed::__Sealed {
     /// The type coercion (`... on SomeType`) applied by the query at this vertex, if any.
     fn coerced_to_type(&self) -> Option<&Arc<str>>;
 
+    fn required_properties(&self) -> Vec<RequiredProperty>;
```
era, line 37: Is this a good signature/name?

trustfall_core/src/interpreter/hints/vertex_info.rs
obi1kenobi: Do you foresee a situation where some optimization would want to know the type of the property as well as its name?

If you can't think of any right now, I think it's fine to not add property types right now but "leave room" for adding them later if needed. An example of leaving room is making the struct non-exhaustive so that such additions aren't major changes.

trustfall_core/test_data/tests/valid_queries/required_properties.graphql.ron
```diff
@@ -0,0 +1,23 @@
...
+            name @output
+        }
+        multiple(max: 5) {
+            name @filter(op: "=", value: ["%name"])
+        }
+
+    }
+
+
+}"#,
```
obi1kenobi, line 19:
```suggestion
        multiple(max: 5) {
            name @filter(op: "=", value: ["%name"])
        }
    }
}"#,
```

trustfall_core/src/interpreter/hints/tests/mod.rs
```diff
@@ -580,6 +580,31 @@ mod static_property_values {
...
+            on_starting_vertices: btreemap! {
+                vid(1) => TrackCalls::<ResolveInfoFn>::new_underlying(Box::new(|info| {
+                    assert!(info.coerced_to_type().is_none());
+                    assert_eq!(vid(1), info.vid());
+
+                    assert_eq!(vec![RequiredProperty::new("value".into()),
+                                    RequiredProperty::new("multiple".into()),
+                                    RequiredProperty::new("predecessor".into()),
+                                    RequiredProperty::new("__typename".into())
+                                ], info.required_properties());
```
obi1kenobi, line 597: Heads-up that this check might break if the order of data returned by `required_properties()` ever changes.

This is a great thing to consider and then explicitly document in the doc comment of that trait method.

trustfall_core/src/interpreter/hints/vertex_info.rs
era: > Is Vec or impl Iterator better?

I think `impl Iterator` would be better here since I don't think the caller needs a specific data structure besides the ability to iterate over it. But I was unsure about it because I will have to add some indirection (Box<dyn object>) since I think we don't want this to be generic, right? Should I change the signature to return `Box<dyn Iterator<Item=RequiredProperty>>`?

> Can the same property appear more than once?
 My understanding of the IR is that it cannot, but I'm not enforcing it. Do you think is worth having a `HashSet` here?
 
 > Is there a guaranteed order in which elements appear?

If I understood correctly, it should because "Trustfall is a deterministic compiler; any non-determinism is considered a bug
and will be caught by tests.". Though I'm not sure if this is meant to be just the output of the compiler (the IR) or everything inside Trustfall. 

But given that we are iterating over a `BTree` for all properties and we are using a `Vec`, we should always produce the same output for the same query.

> Are these only the properties from "this" vertex (the one whose VertexInfo this is, as evidenced by the .vid() accessor), or are they properties from more than one vertex?

I believe it should be only from "this" vertex. Although for graphql it may make sense to also get the properties from the "nested" vertex I thought it would be weird to do such a thing, since my understanding of trustfall is that those vertexes will also be "visited" and the adapter will have to fetch that data.

trustfall_core/src/interpreter/hints/tests/mod.rs
```diff
@@ -580,6 +580,31 @@ mod static_property_values {
...
+        let input_name = "required_properties";
+
+        let adapter = TestAdapter {
+            on_starting_vertices: btreemap! {
+                vid(1) => TrackCalls::<ResolveInfoFn>::new_underlying(Box::new(|info| {
+                    assert!(info.coerced_to_type().is_none());
+                    assert_eq!(vid(1), info.vid());
+
+                    assert_eq!(vec![RequiredProperty::new("value".into()),
+                                    RequiredProperty::new("multiple".into()),
```
era, line 594: I think it makes sense to return `multiple` here because we are folding it, but if we were not using `@fold` I think we should exclude it, correct @obi1kenobi ? I need to confirm it works as intended and fix it.

trustfall_core/src/interpreter/hints/vertex_info.rs
era: Not from the top of my head, so "leaving room" for now

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -132,6 +146,57 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
...
+                .collect::<Vec<RequiredProperty>>(),
+        );
+
+        properties.extend(
+            current_component
+                .folds
+                .values()
+                .map(|i| RequiredProperty::new(i.edge_name.clone()))
+                .collect::<Vec<RequiredProperty>>(),
+        );
```
obi1kenobi, line 178: Edges != properties, those are two disjoint sets of fields. I believe there's already a function for edges in the API.

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -132,6 +145,57 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
         }
     }
 
+    /// required_properties returns an iterator over all properties needed
+    /// for this vertex.
+    fn required_properties(&self) -> Vec<RequiredProperty> {
+        let current_component = self.current_component();
```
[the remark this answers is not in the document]
obi1kenobi: `current_component` here is fine, but make sure that, for example, the properties used for outputs actually belong to the current vertex specifically and not just in general to this component.

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -132,6 +145,57 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
...
+
+        properties.extend(
+            current_component
+                .folds
+                .values()
+                .map(|i| RequiredProperty::new(i.edge_name.clone()))
+                .collect::<Vec<RequiredProperty>>(),
+        );
+
+        // properties.extend(
```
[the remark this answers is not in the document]
obi1kenobi: I think this is best going to be clarified by writing a few more test cases and seeing what the API returns vs what makes sense to return.

trustfall_core/src/interpreter/hints/tests/mod.rs
obi1kenobi: `multiple` is an edge, not a property, so it shouldn't be returned.

Only `__typename, name, value` are properties used at Vid 1 in the test query. `predecessor` and `multiple` are edges, not properties.

Properties are fields whose type is scalar or list of scalar (or list of list of scalar etc.). You can't use a `{` next to a property, there's nothing to expand into. Edges are fields whose type is an interface or object type, or list of such, and they are almost always used with `{` next to them.

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -132,6 +146,58 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
...
+                .filter(|f| match f.right() {
+                    Some(Argument::Tag(FieldRef::ContextField(ctx))) => {
+                        current_vertex.vid == ctx.vertex_id
+                    }
+                    Some(Argument::Tag(FieldRef::FoldSpecificField(fsf))) => {
+                        current_vertex.vid == fsf.fold_root_vid
+                    }
+                    _ => false,
+                })
+                .map(|f| f.right().unwrap().as_tag().unwrap().field_name().into())
```
era, line 193: TODO: improve this line

trustfall_core/src/interpreter/hints/vertex_info.rs
era: > What happens if the same property has both an output and a filter? Or also a tag?

Ah, good point! I will have to handle the duplicates!

trustfall_core/src/interpreter/hints/tests/mod.rs
```diff
@@ -580,6 +580,30 @@ mod static_property_values {
         assert_eq!(adapter.on_starting_vertices.borrow()[&vid(1)].calls, 1);
     }
 
+    #[test]
```
era, line 583: missing tests:
- list with fold 
- property with filter and output

trustfall_core/src/interpreter/hints/vertex_info.rs
obi1kenobi: Using `.filter_map()` to combine the pattern-match with the mapping might be the way to go here.

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -194,7 +169,24 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
...
-        properties
+        properties.extend(current_component.vertices.values().flat_map(|v| {
+            v.filters
+                .iter()
+                .filter(|f| match f.right() {
+                    Some(Argument::Tag(FieldRef::ContextField(ctx))) => {
+                        current_vertex.vid == ctx.vertex_id
+                    }
+                    Some(Argument::Tag(FieldRef::FoldSpecificField(fsf))) => {
+                        current_vertex.vid == fsf.fold_root_vid
```
obi1kenobi, line 181: This is a good question: should we ever return fold-specific fields like count in `required_property`? (And whichever way we decide, we should have a test case to back it up.)

My gut feeling is "no" because they aren't actually properties on the vertex, but derived values based on edges of the vertex.

What do you think?

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -194,7 +169,24 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
...
+                    }
+                    _ => false,
+                })
+                .map(|f| f.right().unwrap().as_tag().unwrap().field_name().into())
+                .map(RequiredProperty::new)
+                .collect::<Vec<RequiredProperty>>()
+        }));
+
+        Box::new(properties.into_iter())
```
obi1kenobi, line 190: For now, this is totally fine. But it might be nice before merging to make this more lazy: we can avoid materializing everything into a `Vec` and instead chain the iterators (ones before `.collect()`) together and return that chained iterator.

trustfall_core/src/interpreter/hints/tests/mod.rs
era: Thanks! That totally makes sense.

trustfall_core/src/interpreter/hints/vertex_info.rs
era: wow, this is amazing! Thank you!

trustfall_core/src/interpreter/hints/vertex_info.rs
era: I agree with you, removing it

trustfall_core/src/interpreter/hints/vertex_info.rs
era: Ah, that is super cool! Thanks for the tip, changing it to use chain

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -132,6 +147,46 @@ impl<T: InternalVertexInfo + super::sealed::__Sealed> VertexInfo for T {
         }
     }
 
+    /// required_properties returns an iterator over all properties needed
```
era, line 150: TODO: docs

trustfall_core/src/interpreter/hints/tests/mod.rs
```diff
@@ -580,6 +581,31 @@ mod static_property_values {
...
+            on_starting_vertices: btreemap! {
+                vid(1) => TrackCalls::<ResolveInfoFn>::new_underlying(Box::new(|info| {
+                    assert!(info.coerced_to_type().is_none());
+                    assert_eq!(vid(1), info.vid());
+
+                    assert_eq!(hashset![
+                                    RequiredProperty::new("name".into()),
+                                    RequiredProperty::new("value".into()),
+                                    RequiredProperty::new("__typename".into())
+                                ], info.required_properties().collect::<HashSet<RequiredProperty>>());
```
obi1kenobi, line 598: nit: can we collect into a `Vec` here to ensure we're also testing that the order of items remains unchanged?

I think for _external_ purposes, we shouldn't promise or document any particular order. But for our own test purposes, I think we want to stick to determinism so that snapshot tests remain easy and we don't get any flaky failures.

obi1kenobi (commented):
Doing great! 🎉 I think just docs and tests left at this point.

trustfall_core/src/interpreter/hints/vertex_info.rs
era: Hum... I'm not sure where else this would better fit, what do you think @obi1kenobi ?

obi1kenobi: I think it's fine here. Just make sure to re-export it wherever `VertexInfo` itself is re-exported, including in the `trustfall` crate?

era: Ah, I totally missed that! Exported it on those places

trustfall_core/test_data/tests/valid_queries/required_properties_list_fold.graphql.ron
```diff
@@ -0,0 +1,15 @@
+TestGraphQLQuery (
+    schema_name: "numbers",
+    query: r#"
+{
+    Number(min: 1, max: 2) {
+        __typename @filter(op: "=", value: ["$type"])
+        value @output @filter(op: "=", value: ["$val"])
+        vowelsInName @output @fold @transform(op: "count")
```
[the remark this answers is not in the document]
era: Created: https://github.com/obi1kenobi/trustfall/issues/450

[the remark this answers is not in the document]
obi1kenobi: Ah yikes, `vowelsInName` is a property, not an edge, so even `@fold` shouldn't be allowed there 😅

Great catch on both! Wanna open another issue for that, or amend the existing one?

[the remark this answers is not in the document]
era: So the only way to trigger the bug is to have both: `@fold` and `@transform`, without it you have the right behaviour:
`Unexpected error: @transform applied to non-folded edge field` so I think the example for the issue still holds, but I can add this observation for completeness.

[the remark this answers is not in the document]
era: Ah, but `@fold` allow also triggers the bug hahaha, I will amend the issue with that, thanks!

trustfall_core/src/interpreter/hints/vertex_info.rs
```diff
@@ -23,6 +37,17 @@ pub trait VertexInfo: super::sealed::__Sealed {
     /// The type coercion (`... on SomeType`) applied by the query at this vertex, if any.
     fn coerced_to_type(&self) -> Option<&Arc<str>>;
 
+    /// Return all properties required for the current vertex, including: output, filtered, and
+    /// tagged properties. It's guaranteed that each property will only show once in the iterator,
+    /// so even if a property is been used as a filter and output, it will only show once.
```
obi1kenobi, line 42:
```suggestion
    /// so even if a property has been used as a filter and output, it will only show once.
```

obi1kenobi (approved):
Nicely done!

Issue #450, opened 2023-08-18
Query accepting properties with @transform


era:
Trustfall currently doesn't support using `@transform` on properties, but we don't have a validation in-place to avoid users doing such query. For example using the [Trustfall Playground](https://play.predr.ag/hackernews#?f=2&q=*3-Latest-stories-on-HackerNews-that-have-a-minimum-score-and-karma*lquery---0Latest*9max*B-100*0---2title-*o*l--_2byUsername-*o*l--_2submittedUrl-*o*l--_2score-*o-*f*9*p-***G*e***L-*v-*c***4minScore***j*0*l--_2storyUrl*B-url-*o-*F-*T*9*p-**count***0*l*l--_2byUser---4karma-*o-*f*9*p-***G*e***L-*v-*c***4minKarma***j*0*l--_2--*0*J*l*J&v=--0**minScore***B-5*L*l--_0**minKarma***B-300*l*J) is possible to run:

```
# Latest stories on HackerNews that have a minimum score and karma
query {
  Latest(max: 100) {
    title @output
    byUsername @output
    submittedUrl @output
    score @output @filter(op: ">=", value: ["$minScore"])
    storyUrl: url @output @fold @transform(op: "count")

    byUser {
      karma @output @filter(op: ">=", value: ["$minKarma"])
    }
  }
}
```

Which returns:

```
[
  {
    "byUsername": "thunderbong",
    "karma": 54591,
    "score": 26,
    "storyUrl": "https://news.ycombinator.com/item?id=37175721",
    "submittedUrl": "https://fortomorrow.org/explore-solutions/sargablock",
    "title": "Sargablock: Bricks from Seaweed"
  }
]
```
Note that `@fold` is also not allowed on a property. So the following example should not be allowed as well:

```
# Latest stories on HackerNews that have a minimum score and karma
query {
  Latest(max: 100) {
    title @output
    byUsername @output
    submittedUrl @output
    score @output @filter(op: ">=", value: ["$minScore"])
    storyUrl: url @output @fold

    byUser {
      karma @output @filter(op: ">=", value: ["$minKarma"])
    }
  }
}
```

obi1kenobi:
Ah, whoops! Nice catch.

PR #457, opened 2023-08-21
Does not allow properties to use fold

era:
Fix https://github.com/obi1kenobi/trustfall/issues/450

Validates on the frontend if the query is doing `@fold` on a property.

### DIFF
--- a/trustfall_core/src/frontend/error.rs
+++ b/trustfall_core/src/frontend/error.rs
@@ -109,6 +109,12 @@ pub enum FrontendError {
     #[error("The query failed to validate against the schema: {0}")]
     ValidationError(#[from] ValidationError),
 
+    #[error(
+        "The field \"{0}\" is a property but the query uses @fold.\
+     Only edges can be used with @fold."
+    )]
+    FoldOnProperty(String),
+
     #[error("Unexpected error: {0}")]
     OtherError(String),
 }
--- a/trustfall_core/src/frontend/mod.rs
+++ b/trustfall_core/src/frontend/mod.rs
@@ -1009,7 +1009,6 @@ where
             subfield_post_coercion_type,
             subfield_raw_type,
         ) = get_field_name_and_type_from_schema(defined_fields, subfield);
-
         if schema.vertex_types.contains_key(subfield_post_coercion_type.as_ref()) {
             // Processing an edge.
 
@@ -1103,6 +1102,11 @@ where
             || subfield_name.as_ref() == TYPENAME_META_FIELD
         {
             // Processing a property.
+
+            // @fold is not allowed on a property
+            if let Some(fold_group) = &connection.fold {
+                errors.push(FrontendError::FoldOnProperty(subfield.name.to_string()));
+            }
 
             let subfield_name: Arc<str> = subfield_name.as_ref().to_owned().into();
             let key = (current_vid, subfield_name.clone());
--- a/trustfall_core/test_data/tests/frontend_errors/transform_used_on_property.frontend-error.ron
+++ b/trustfall_core/test_data/tests/frontend_errors/transform_used_on_property.frontend-error.ron
@@ -1,0 +1,1 @@
+Err(FoldOnProperty("vowelsInName"))
--- a/trustfall_core/test_data/tests/frontend_errors/fold_used_on_property.frontend-error.ron
+++ b/trustfall_core/test_data/tests/frontend_errors/fold_used_on_property.frontend-error.ron
@@ -1,0 +1,1 @@
+Err(FoldOnProperty("vowelsInName"))
--- a/trustfall_core/test_data/tests/frontend_errors/fold_used_on_property.graphql-parsed.ron
+++ b/trustfall_core/test_data/tests/frontend_errors/fold_used_on_property.graphql-parsed.ron
@@ -1,0 +1,53 @@
+Ok(TestParsedGraphQLQuery(
+  schema_name: "numbers",
+  query: Query(
+    root_connection: FieldConnection(
+      position: Pos(
+        line: 3,
+        column: 5,
+      ),
+      name: "Two",
+    ),
+    root_field: FieldNode(
+      position: Pos(
+        line: 3,
+        column: 5,
+      ),
+      name: "Two",
+      connections: [
+        (FieldConnection(
+          position: Pos(
+            line: 4,
+            column: 9,
+          ),
+          name: "value",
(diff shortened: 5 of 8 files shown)


trustfall_core/test_data/tests/frontend_errors/transform_used_on_property.frontend-error.ron
```diff
@@ -0,0 +1 @@
+Err(FoldOnProperty("vowelsInName"))
```
era, line 1: Using `@transform` without `@fold` is caught at another place in the code. And because we are not allowing `@fold` on the property, the whole usage of `@transform` is blocked for properties.

trustfall_core/src/frontend/error.rs
```diff
@@ -109,6 +109,12 @@ pub enum FrontendError {
     #[error("The query failed to validate against the schema: {0}")]
     ValidationError(#[from] ValidationError),
 
+    #[error(
+        "The field \"{0}\" is a property but the query uses @fold.\
+     Only edges can be used with @fold."
+    )]
+    FoldOnProperty(String),
+
```
obi1kenobi, line 117: A few questions here. I genuinely don't know the answer to these off the top of my head, but we probably want to consider them so as to make consistent and intuitive choices:
- What error does the frontend raise for other edge-only directives (`@optional` / `@recurse`) is used on a property? (Not sure if we already have test cases for these, but we should add them if we don't!)
- Would it make sense to use a more general variant for all such directives that are invalid on properties, like a `UnsupportedDirectiveOnProperty`? (Name subject to workshopping, open to ideas.)

Also, a couple of nits:
- We are not guaranteeing enum variant order, so let's try to group this variant near other variants for related errors instead.
- The error message is a bit unclear: for example, it only says that `@fold` was used by never says the actual problem — that it was used by applying it _to a property_. The existing error messages here are also not good, but good diagnostics are super important so we should try to start raising the quality bar if at all possible.

trustfall_core/test_data/tests/frontend_errors/transform_used_on_property.frontend-error.ron
obi1kenobi: Would you mind renaming this test case to `fold_and_transform_used_on_property` to avoid confusion?

In principle, `@transform` by itself on properties should be valid, and I'd like to support it in the future. We just currently don't have any implemented property transformations, so `@transform` by itself on properties is therefore also invalid because it's useless.

Here renaming the test case is "leaving room" for `@transform` on properties to have a test case take this name which is more appropriate for it.

obi1kenobi (commented):
Nice work! Uncovered a bit of a design question in the process, which I'm sure should be possible to overcome without too much trouble.

trustfall_core/src/frontend/error.rs
era: > What error does the frontend raise for other edge-only directives (`@optional` / `@recurse`) is used on a property? (Not sure if we already have test cases for these, but we should add them if we don't!)

Unless the playground is outdated and validations were added after, those do not raise any errors. So it feels like this change should include those directives as well and use naming like you proposed (`UnsupportedDirectiveOnProperty`)

> We are not guaranteeing enum variant order, so let's try to group this variant near other variants for related errors instead.
> The error message is a bit unclear: for example, it only says that `@fold` was used by never says the actual problem — that it was used by applying it to a property. The existing error messages here are also not good, but good diagnostics are super important so we should try to start raising the quality bar if at all possible.

I will send out a version with a little better wording, but I will sleep on it to see if I can improve it.

obi1kenobi: I don't think any validations have been added in a while. Oops! Great catch, thanks for finding this!

trustfall_core/src/frontend/mod.rs
```diff
@@ -1104,6 +1103,14 @@ where
         {
             // Processing a property.
 
+            // @fold is not allowed on a property
+            if let Some(fold_group) = &connection.fold {
```
era, line 1107: todo: add other directives

trustfall_core/src/frontend/error.rs
```diff
@@ -31,6 +31,11 @@ pub enum FrontendError {
     )]
     TagUsedOutsideItsFoldedSubquery(String, String),
 
+    #[error(
+        "It's not allowed to apply \"{0}\" to a property, but it was applied to \"{1}\" property ."
+    )]
+    UnsupportedDirectiveOnProperty(String, String),
```
obi1kenobi, line 37: As another option for placing the variant, consider having it next to `UnsupportedEdgeOutput`. That variant is about an unsupported directive on an edge, which nicely parallels this variant's complaint about an unsupported directive on a property.

trustfall_core/src/frontend/mod.rs
```diff
@@ -1104,6 +1103,30 @@ where
         {
             // Processing a property.
 
+            // @fold is not allowed on a property
+            if connection.fold.is_some() {
```
era, line 1107: There is some code duplication around here, but at least the signatures I thought about would keep things highly coupled with the context of this method. So I kept the code duplication because I thought it was the cleaner version of the code. Let me know if you have ideas for making this better.

obi1kenobi: This is totally fine. Not all duplication is bad, and in this case the duplicated code is fairly short and very clear. Anything less duplicated would be way more convoluted and less readable.

trustfall_core/src/frontend/error.rs
```diff
@@ -52,6 +52,9 @@ pub enum FrontendError {
     #[error("Incompatible types encountered in @filter: {0}")]
     FilterTypeError(#[from] FilterTypeError),
 
+    #[error("Found {0} applied to \"{1}\" property, this directive can only be applied to edges.")]
```
obi1kenobi, line 55: Minor stylistic suggestion:
```suggestion
    #[error("Found {0} applied to \"{1}\" property, which is not supported since that directive can only be applied to edges.")]
```

obi1kenobi (commented):
This looks great, I'd be happy to merge as soon as the fix for that one test case is in 🎉

obi1kenobi:
Also remember to update the PR title and description to include the other directives being handled!

obi1kenobi:
Ah the failing lints are because we got a new Rust stable today. I'll push a fix real quick, sorry about that.

obi1kenobi (approved):

PR #461, opened 2023-08-27
Add better filter error message for string instead of array.

u9g:
This is a common mistake for me so I can only imagine how many beginners fall into this trap, let's give a gentler error message for this case.

Example of the new error message: `Directive @filter received a string instead of a list. Did you mean to put '@filter(op: "!=", value: ["$empty_string"])' instead?`

### DIFF
--- a/trustfall_core/src/graphql_query/error.rs
+++ b/trustfall_core/src/graphql_query/error.rs
@@ -24,6 +24,12 @@ pub enum ParseError {
     #[error("Directive {0} received value of inappropriate type for argument {1}")]
     InappropriateTypeForDirectiveArgument(String, String, Pos),
 
+    #[error(
+        "Directive @filter received a string instead of a list. \
+        Did you mean to put '@filter(op: \"{0}\", value: [\"{1}\"])' instead?"
+    )]
+    FilterExpectsListNotString(String, String, Pos),
+
     #[error("Field {0} received an invalid value for argument {1}: {2}")]
     InvalidFieldArgument(String, String, Value, Pos),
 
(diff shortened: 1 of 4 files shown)


trustfall_core/src/graphql_query/error.rs
```diff
@@ -24,6 +24,12 @@ pub enum ParseError {
     #[error("Directive {0} received value of inappropriate type for argument {1}")]
     InappropriateTypeForDirectiveArgument(String, String, Pos),
 
+    #[error(
+        "Directive @filter received a string instead of a list. \
```
u9g, line 28: This doesn't really fit the style, a more fitting error message to the ones below would be `Directive @filter received a string, this is not supported.` However I feel that as this error also gives a possible solution, I think we should tell them what the right answer would be

obi1kenobi: I think the new style is good here. Just one issue: "received" a string / list doesn't sound right because `@filter` takes two args, not one. Any way to tweak the wording to make this clearer?

u9g: `Directive @filter's second argument is a string instead of a list.` OR
`Directive @filter's value argument is a string instead of a list.` (I prefer second)

obi1kenobi: Perhaps `Value argument in @filter directive is a string instead of a list.` ?

u9g: Yep, sounds great, I was sticking to the existing style but I'm good with this one too, I'll make the change.

obi1kenobi (approved):

Issue #415, opened 2023-07-29
Change CandidateValues to no longer return a reference to a FieldValue


u9g:
This was a good optimization when FieldValues held their own values, now they hold Arc's of their own values so it's no longer as justifiable. Returning a reference to a FieldValue complicates lifetimes of everything that touches them, and especially in adapters that may want to use them, this gets annoying quick.

PR #466, opened 2023-08-29
Remove reference from statically_required_property's FieldValue

u9g:
closes https://github.com/obi1kenobi/trustfall/issues/415

### DIFF
--- a/trustfall_core/src/interpreter/hints/candidates.rs
+++ b/trustfall_core/src/interpreter/hints/candidates.rs
@@ -30,7 +30,7 @@ pub enum CandidateValue<T> {
 
 impl<T: Clone> CandidateValue<&T> {
     /// Converts from `CandidateValue<&T>` to `CandidateValue<T>`.
-    pub fn cloned(&self) -> CandidateValue<T> {
+    pub(crate) fn cloned(&self) -> CandidateValue<T> {
         match self {
             CandidateValue::Impossible => CandidateValue::Impossible,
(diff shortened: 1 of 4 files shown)


trustfall_core/src/interpreter/hints/candidates.rs
```diff
@@ -28,19 +28,6 @@ pub enum CandidateValue<T> {
     All,
 }
 
-impl<T> CandidateValue<T> {
```
u9g, line 31: We don't want to return back `CandidateValue<&T>` anymore so this isn't necessary anymore

obi1kenobi: I think either leave `.as_ref()` or also remove `.cloned()` in the impl below. They are duals of each other.

u9g: I can't remove `.cloned()` as we still internally use `CandidateValue<&FieldValue>`, but before we return it, we call `.cloned()` to remove the reference, see: https://github.com/obi1kenobi/trustfall/pull/466/files#diff-84f518e6512bfe60e05ad1f0c0fe9a9f63e4e0220a652aa15d07c1af6a4e24f1R229

u9g: I did just make it pub(crate) though, as we should be the only consumers of this API after this change.

obi1kenobi: sounds good!

obi1kenobi (approved):

PR #481, opened 2023-10-07
Allow adapters to process types that deref to their vertex type.

obi1kenobi:
Before this PR, adapter implementations were unnecessarily strict: they required that the `DataContext` contain _specifically_ their `Vertex` type. They would refuse to work in a situation where the `DataContext` contained a type that _within it_ contained a `Vertex` i.e. could dereference to a `Vertex` upon request. That made it harder than necessary to expose adapters across programming languages, or use them as building blocks that can combine their functionality.

This PR changes the `Adapter` trait to allow any `DataContext<V>` representation, so long as that `V` type implements `AsVertex<Self::Vertex>` and has an appropriate lifetime for the adapter. When the adapter needs to look up the `active_vertex()` and get an `Option<Self::Vertex>` back, it can do so via the `AsVertex` trait which handles the conversion. `AsVertex` has a blanket impl for all `Debug + Clone` (i.e. valid `Adapter::Vertex`) types, so implementations should be able to upgrade to the new API by just changing the `Adapter` fn signatures and without implementation changes.


trustfall_core/src/interpreter/mod.rs
```diff
@@ -92,8 +92,65 @@ impl<Vertex> DataContext<Vertex> {
...
-    pub fn active_vertex(&self) -> Option<&Vertex> {
-        self.active_vertex.as_ref()
+    pub fn active_vertex<V>(&self) -> Option<&V>
+    where
+        Vertex: AsVertex<V>,
+    {
+        self.active_vertex.as_ref().and_then(|v| v.as_vertex())
+    }
+
+    pub fn map<T>(self, mut mapper: impl FnMut(Vertex) -> T) -> DataContext<T> {
```
obi1kenobi, line 102: pub fn needs docs

trustfall_core/src/interpreter/mod.rs
```diff
@@ -92,8 +92,65 @@ impl<Vertex> DataContext<Vertex> {
...
+                .collect(),
+            folded_values: self.folded_values,
+            piggyback: self
+                .piggyback
+                .map(|v| v.into_iter().map(|ctx| ctx.map(&mut mapper)).collect()),
+            imported_tags: self.imported_tags,
+        }
+    }
+
+    pub fn flat_map<T>(self, mut mapper: impl FnMut(Vertex) -> Option<T>) -> DataContext<T> {
```
obi1kenobi, line 127: pub fn needs docs

trustfall_core/src/interpreter/mod.rs
```diff
@@ -523,11 +580,31 @@ pub trait Adapter<'vertex> {
...
+
+pub trait AsVertex<V>: Debug + Clone {
+    fn as_vertex(&self) -> Option<&V>;
+
+    fn into_vertex(self) -> Option<V>;
+}
+
+pub trait Cast<V>: AsVertex<V> {
+    fn into_self(vertex: V) -> Self;
+}
```
obi1kenobi, line 600: need docs for all of these

trustfall_core/src/interpreter/mod.rs
```diff
@@ -446,13 +503,13 @@ pub trait Adapter<'vertex> {
     /// - Produce contexts in the same order as the input `contexts` iterator produced them.
     /// - Produce property values whose type matches the property's type defined in the schema.
     /// - When a context's active vertex is `None`, its property value is [`FieldValue::Null`].
-    fn resolve_property(
+    fn resolve_property<V: AsVertex<Self::Vertex> + 'vertex>(
```
obi1kenobi, line 506: Need docs updates for all the trait methods here.

trustfall_core/src/interpreter/trace.rs
```diff
@@ -196,6 +196,11 @@ where
     })
 }
 
+// Tapping adapters must be done at top level, where the `<V>` generic on the resolver methods
+// is the same as `AdapterT::Vertex`.
+//
+// Otherwise, the recorded traces may not be possible to replay since they would be incomplete:
+// they only capture a portion of the execution, the rest of which is missing.
```
obi1kenobi, line 203: This should probably be a doc-comment `///` on `AdapterTap`

PR #490, opened 2023-10-16
Improve patches on graphql-language-service and remove monaco-editor patch

u9g:
I remove the `monaco-editor` patch because it was the wrong fix, and we can just wait for the pr to land as it's just a console message. 

This pr adds:

- autocompleting fields on ctrl+space (on a new line)
- autocompleting @-directives (after a field name or more @directives) and will somewhat filter them based on the field's type
- Only allow interfaces to be autocompleted that implement the type in `... on X` fragments
- Updates pnpm because they retroactively broke our pnpm version for patches as dependency installs fail and the solution [update to latest](https://github.com/pnpm/pnpm/issues/6424#issuecomment-1515748464)

Future possibilities:
- Autocomplete `op` type in `@filter`?


experiments/browser_based_querying/patches/graphql-language-service@5.1.7.patch
```diff
@@ -1,21 +1,51 @@
...
 --- a/esm/interface/getAutocompleteSuggestions.js
 +++ b/esm/interface/getAutocompleteSuggestions.js
-@@ -234,7 +234,12 @@ export function getAutocompleteSuggestions(schema, queryText, cursor, contextTok
+@@ -63,6 +63,9 @@ const hasTypeSystemDefinitions = (sdl) => {
+     }
+     return hasTypeSystemDef;
+ };
++const primitives = new Set(['Int', 'Float', 'String', 'Boolean'])
++const propertyDirectives = new Set(['fold', 'recurse', 'optional', 'transform'])
++const edgeDirectives = new Set(['filter', 'tag', 'output', 'transform'])
```
obi1kenobi, line 11: These don't seem like they are in the right categories. Are the names `propertyDirectives` and `edgeDirectives` backwards?

experiments/browser_based_querying/patches/graphql-language-service@5.1.7.patch
```diff
@@ -1,21 +1,51 @@
...
      if (kind === RuleKinds.DIRECTIVE) {
 -        return getSuggestionsForDirective(token, state, schema, kind);
 +        // We should autocomplete fields on the line after we autocomplete a directive
-+        if (state.needsAdvance) { // active after '@' or many other kinds of punctuation, so perfect for when we need to autocomplete directives
-+            return getSuggestionsForDirective(token, state, schema, kind);
++        if (state.needsAdvance || state.name === null) { // active after '@' or many other kinds of punctuation, so perfect for when we need to autocomplete directives.
++                                                         // `state` is `null` when the user autocompletes after typing just '@' then opening autocomplete menu
++            // further restrict the directives we suggest based on the type of the field we are on
++            // x.label does not include the '@'
++            if (primitives.has(typeInfo.fieldDef.type.name)) {
```
obi1kenobi, line 25: What happens here if the field type is something like `[String!]` or `[[Int]]!`?

obi1kenobi (commented):
Love the ideas for smarter autocomplete! Thanks for putting this together.

I think I might have found a few remaining edge cases before this is ready to merge and publish. Whenever you're able to take a look, I'll be happy to merge this.

experiments/browser_based_querying/patches/graphql-language-service@5.1.7.patch
u9g: Amazing point, thank you! Then it will be an object of

```json
{
    "ofType": { // <- the list
        "name": "String" // <- the actual type
    }
}
```

So the solution is

experiments/browser_based_querying/pnpm-lock.yaml
```diff
@@ -1,12 +1,13 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
```
obi1kenobi, line 4: Do we really want this setting?

experiments/browser_based_querying/patches/graphql-language-service@5.1.7.patch
u9g: [bb3fb60](https://github.com/obi1kenobi/trustfall/pull/490/commits/bb3fb604678ab10f2d039088d716e28c0677bd41)

PR #501, opened 2023-10-22
Use GitHub permalinks for Adapter docs.

obi1kenobi:


nvzqz:
I recommend instead linking to the tag of a specific Trustfall version.

obi1kenobi:
Of course. No version containing these is currently out, hence pointing to a commit instead.

obi1kenobi:
> I recommend instead linking to the tag of a specific Trustfall version.

Done in #517.

PR #524, opened 2023-12-06
Allow new Safari and iOS in the playground.

obi1kenobi:
The two bugs blocking the HackerNews playground from working on Safari and iOS appear to have been fixed in a recent update. I've confirmed that the playground now works on Safari 16.6 on both desktop and iOS. Note that iOS 16.7 is the one that apparently ships with Safari 16.6, based on the device I was able to check with.


obi1kenobi:
I'm going to merge this since it's an improvement over what we have right now, but I'm happy to make further improvements in separate PRs if anything comes up in code review :)

PR #540, opened 2024-01-18
Add information to unimplemented! messages

fprasx:
The first time I ran `trustfall_stubgen` this is what I saw:
```
thread 'main' panicked at~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trustfall_core-0.7.1/src/schema/mod.rs:163:46:
not implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This should alleviate that a bit. At some point, we should also have a reference detailing the difference between Trustfall and GraphQl schemas.

### DIFF
--- a/trustfall_core/src/schema/mod.rs
+++ b/trustfall_core/src/schema/mod.rs
@@ -124,7 +124,7 @@ directive @transform(op: String!) on FIELD
                 TypeSystemDefinition::Schema(s) => {
                     assert!(schema.is_none());
                     if s.node.extend {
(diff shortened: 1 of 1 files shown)


obi1kenobi (approved):
Thanks! Yeah, all of this is stuff we should support _someday_ but isn't a high priority right now, since Trustfall is quite usable without it.

fprasx:
Pushed a format, should pass now

obi1kenobi (approved):

PR #595, opened 2024-04-16
feat/594: add `VariableDefinitionInQuery` parse error

devanbenz:
- Adds `VariableDefinitionInQuery` parse error


trustfall_core/test_data/tests/parse_errors/variable_in_root_query.graphql.ron
```diff
@@ -0,0 +1,11 @@
+TestGraphQLQuery (
+    schema_name: "numbers",
+    query: r#"{
+  query Number($max: Int!) {
```
devanbenz, line 4: @obi1kenobi I can't seem to figure out how to write a graphql query to test this 😅 
```
query HeroForEpisode($ep: Episode!) {
  hero(episode: $ep) {
    name
    ... on Droid {
      primaryFunction
    }
    ... on Human {
      height
    }
  }
}
```
Using the following example ^ This will throw a `MultipleOperationsInDocument` before it can throw a `VariableDefinitionInRootQuery` and I can't seem to find a way to use variables in graphql outside of having a top level query definition like so. I'm assuming trustfall uses graphql but as a different dialect where this is not allowed in the first place for queries.

obi1kenobi: Huh. I'm confused too, I don't understand what the multiple definitions are. That sounds like a bug to me, and it's possibly in my understanding of some GraphQL core concept.

Could you `Debug` print the multiple definitions and put them here?

I would also expect that query to trigger the `VariableDefinitionInRootQuery` error case in the parser. Feel free to tweak the logic to make it so, unless the multiple definitions thing seems to indicate we'd be making a mistake doing that.

devanbenz:
```
{Name("MyQuery"): Positioned { pos: Pos(2:1), node: OperationDefinition { ty: Query, variable_definitions: [Positioned { pos: Pos(2:15), node: VariableDefinition { name: Positioned { pos: Pos(2:16), node: Name("max") }, var_type: Positioned { pos: Pos(2:21), node: Type { base: Named(Name("Int")), nullable: false } }, directives: [], default_value: None } }], directives: [], selection_set: Positioned { pos: Pos(2:27), node: SelectionSet { items: [Positioned { pos: Pos(3:5), node: Field(Positioned { pos: Pos(3:5), node: Field { alias: None, name: Positioned { pos: Pos(3:5), node: Name("Number") }, arguments: [(Positioned { pos: Pos(3:12), node: Name("max") }, Positioned { pos: Pos(3:17), node: Variable(Name("max")) })], directives: [], selection_set: Positioned { pos: Pos(3:23), node: SelectionSet { items: [Positioned { pos: Pos(4:9), node: Field(Positioned { pos: Pos(4:9), node: Field { alias: Some(Positioned { pos: Pos(4:9), node: Name("num") }), name: Positioned { pos: Pos(4:14), node: Name("value") }, arguments: [], directives: [Positioned { pos: Pos(4:20), node: Directive { name: Positioned { pos: Pos(4:21), node: Name("output") }, arguments: [] } }], selection_set: Positioned { pos: Pos(0:0), node: SelectionSet { items: [] } } } }) }] } } } }) }] } } } }}
```

This is whats debug printing for the match case where its throwing `MultipleOperationsInDocument` error.  From the graphql async parser: 

```
The operations of a GraphQL document.
There is either one anonymous operation or many named operations.
```

Is the expectation that trustfall should only support anonymous operations? Or that named operations should be parsed out? It's the parse library thats lexing this as multiple operations.

obi1kenobi: Could you debug print `document` as a whole, and use pretty-printed output `{:#?}` to make it easier to read?

I can't tell exactly which part of the query this printed output is, and it's quite difficult to read too because it's all on one gigantic line.

I don't think we have a stance on named vs anonymous operations. I just don't see multiple operations in here at all, and I'm wondering whether that's a bug upstream or a bug in my understanding of the GraphQL spec.

devanbenz: Sorry my bad 😆 - I'll go ahead and pretty print it. 
```
{
    Name(
        "MyQuery",
    ): Positioned {
        pos: Pos(2:1),
        node: OperationDefinition {
            ty: Query,
            variable_definitions: [
                Positioned {
                    pos: Pos(2:15),
                    node: VariableDefinition {
                        name: Positioned {
                            pos: Pos(2:16),
                            node: Name(
                                "max",
                            ),
                        },
                        var_type: Positioned {
                            pos: Pos(2:21),
                            node: Type {
                                base: Named(
                                    Name(
                                        "Int",
                                    ),
                                ),
                                nullable: false,
                            },
                        },
                        directives: [],
                        default_value: None,
                    },
                },
            ],
            directives: [],
            selection_set: Positioned {
                pos: Pos(2:27),
                node: SelectionSet {
                    items: [
                        Positioned {
                            pos: Pos(3:5),
                            node: Field(
                                Positioned {
                                    pos: Pos(3:5),
                                    node: Field {
                                        alias: None,
                                        name: Positioned {
                                            pos: Pos(3:5),
                                            node: Name(
                                                "Number",
                                            ),
                                        },
                                        arguments: [
                                            (
                                                Positioned {
                                                    pos: Pos(3:12),
                                                    node: Name(
                                                        "max",
                                                    ),
                                                },
                                                Positioned {
                                                    pos: Pos(3:17),
                                                    node: Variable(
                                                        Name(
                                                            "max",
                                                        ),
                                                    ),
                                                },
                                            ),
                                        ],
                                        directives: [],
                                        selection_set: Positioned {
                                            pos: Pos(3:23),
                                            node: SelectionSet {
                                                items: [
                                                    Positioned {
                                                        pos: Pos(4:9),
                                                        node: Field(
                                                            Positioned {
                                                                pos: Pos(4:9),
                                                                node: Field {
                                                                    alias: Some(
                                                                        Positioned {
                                                                            pos: Pos(4:9),
                                                                            node: Name(
                                                                                "num",
                                                                            ),
                                                                        },
                                                                    ),
                                                                    name: Positioned {
                                                                        pos: Pos(4:14),
                                                                        node: Name(
                                                                            "value",
                                                                        ),
                                                                    },
                                                                    arguments: [],
                                                                    directives: [
                                                                        Positioned {
                                                                            pos: Pos(4:20),
                                                                            node: Directive {
                                                                                name: Positioned {
                                                                                    pos: Pos(4:21),
                                                                                    node: Name(
                                                                                        "output",
                                                                                    ),
                                                                                },
                                                                                arguments: [],
                                                                            },
                                                                        },
                                                                    ],
                                                                    selection_set: Positioned {
                                                                        pos: Pos(0:0),
                                                                        node: SelectionSet {
                                                                            items: [],
                                                                        },
                                                                    },
                                                                },
                                                            },
                                                        ),
                                                    },
                                                ],
                                            },
                                        },
                                    },
                                },
                            ),
                        },
                    ],
                },
            },
        },
    },
}
```

devanbenz: FYI this is the following document:
```
query MyQuery($max: Int!) {
    Number(max: $max) {
        num: value @output
    }
}
```

Normally when I use graphQL we don't rely on anonymous operations so more often then not its formatted like the above ^^ or like VVV
```
mutation Mutation($vars: Type!) {
	operation(foo: $vars) {
		returnVal
	}
}
```

devanbenz: When printing out the `&document` value before this match statement https://github.com/obi1kenobi/trustfall/blob/f94cab5bd26e64ac095d2d4242edd10b9c9b2c6e/trustfall_core/src/graphql_query/query.rs#L126

I'm seeing:
```
ExecutableDocument {
    operations: Multiple(
        {
            Name(
                "MyQuery",
            ): Positioned {
                pos: Pos(2:1),
                node: OperationDefinition {
                    ty: Query,
                    variable_definitions: [
                        Positioned {
                            pos: Pos(2:15),
                            node: VariableDefinition {
                                name: Positioned {
                                    pos: Pos(2:16),
                                    node: Name(
                                        "max",
                                    ),
                                },
                                var_type: Positioned {
                                    pos: Pos(2:21),
                                    node: Type {
                                        base: Named(
                                            Name(
                                                "Int",
                                            ),
                                        ),
                                        nullable: false,
                                    },
                                },
                                directives: [],
                                default_value: None,
                            },
                        },
                    ],
                    directives: [],
                    selection_set: Positioned {
                        pos: Pos(2:27),
                        node: SelectionSet {
                            items: [
                                Positioned {
                                    pos: Pos(3:5),
                                    node: Field(
                                        Positioned {
                                            pos: Pos(3:5),
                                            node: Field {
                                                alias: None,
                                                name: Positioned {
                                                    pos: Pos(3:5),
                                                    node: Name(
                                                        "Number",
                                                    ),
                                                },
                                                arguments: [
                                                    (
                                                        Positioned {
                                                            pos: Pos(3:12),
                                                            node: Name(
                                                                "max",
                                                            ),
                                                        },
                                                        Positioned {
                                                            pos: Pos(3:17),
                                                            node: Variable(
                                                                Name(
                                                                    "max",
                                                                ),
                                                            ),
                                                        },
                                                    ),
                                                ],
                                                directives: [],
                                                selection_set: Positioned {
                                                    pos: Pos(3:23),
                                                    node: SelectionSet {
                                                        items: [
                                                            Positioned {
                                                                pos: Pos(4:9),
                                                                node: Field(
                                                                    Positioned {
                                                                        pos: Pos(4:9),
                                                                        node: Field {
                                                                            alias: Some(
                                                                                Positioned {
                                                                                    pos: Pos(4:9),
                                                                                    node: Name(
                                                                                        "num",
                                                                                    ),
                                                                                },
                                                                            ),
                                                                            name: Positioned {
                                                                                pos: Pos(4:14),
                                                                                node: Name(
                                                                                    "value",
                                                                                ),
                                                                            },
                                                                            arguments: [],
                                                                            directives: [
                                                                                Positioned {
                                                                                    pos: Pos(4:20),
                                                                                    node: Directive {
                                                                                        name: Positioned {
                                                                                            pos: Pos(4:21),
                                                                                            node: Name(
                                                                                                "output",
                                                                                            ),
                                                                                        },
                                                                                        arguments: [],
                                                                                    },
                                                                                },
                                                                            ],
                                                                            selection_set: Positioned {
                                                                                pos: Pos(0:0),
                                                                                node: SelectionSet {
                                                                                    items: [],
                                                                                },
                                                                            },
                                                                        },
                                                                    },
                                                                ),
                                                            },
                                                        ],
                                                    },
                                                },
                                            },
                                        },
                                    ),
                                },
                            ],
                        },
                    },
                },
            },
        },
    ),
    fragments: {},
}
```

devanbenz: Taking a look at the source code for the graphQL parsing library this is where its defining the enum that gets returned:
https://docs.rs/async-graphql-parser/latest/src/async_graphql_parser/parse/executable.rs.html#30

obi1kenobi: Based on reading the GraphQL spec, it seems like `Single` vs `Multiple` is a misnomer and better names might be `SingleAnonymous` vs `Named`.

With that in mind, I feel our best course of action would be to leave a comment describing this weirdness (for the sake of future readers of this code), and then write a bunch of test cases to pin down both our upstream library's behavior and our own.

We should accept single named operations and run them normally without error. We should not accept multiple named operations — that's where we want to return the "multiple operations" error, I think.

devanbenz: Okay this sounds good. So something like this 
```
query Query() {
	Number(max: 3) {
		num: value @output
	}
}
```

should be fine but something like this 

```
query Query() {
	Number(max: 3) {
		num: value @output
	}
	Number(max: 8) {
		num: value @output
	}
}
```

should error?

obi1kenobi: I think the latter should error for a different reason -- it has the same output name twice, but there's still only one named operation.

```graphql
query Query1() {
  Number(max: 3) {
    num: value @output
  }
}

query Query2() {
  Number(max: 3) {
    value @output
  }
}
```
This has two named operations, and should error with the "multiple operations" error.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,52 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
 
     match &document.operations {
         DocumentOperations::Multiple(mult) => {
-            return Err(ParseError::MultipleOperationsInDocument(mult.values().next().unwrap().pos))
+            return if mult.values().len() > 1 {
+                Err(ParseError::MultipleOperationsInDocument(mult.values().next().unwrap().pos))
+            } else {
+                let node = mult.values().next().unwrap();
+                parse_operation_definition(node)
+            }
```
obi1kenobi, line 133: I wrote this code really sloppily back when I didn't really know what I was doing in Rust. Would you mind making a couple of improvements along the way?

Could we:
- Replace all `.unwrap()` here with `expect()`
- Use an `if let Some()` to extract the single value instead of `unwrap()/expect()`
- If there is more than one value, report the `pos` of the _second_ operation instead of the first

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,52 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
...
-            let root_selection_set = &root_node.selection_set.node;
-            let root_items = &root_selection_set.items;
-            if root_items.len() != 1 {
-                return Err(ParseError::MultipleQueryRoots(root_items[1].pos));
-            }
+    if !root_node.variable_definitions.is_empty() {
+        let first_variable_definition = root_node.variable_definitions.first().unwrap();
+        return Err(ParseError::VariableDefinitionInRootQuery(first_variable_definition.pos));
+    }
+    if !root_node.directives.is_empty() {
```
obi1kenobi, line 152: These should both probably be `if let Some()` as well. Again, this is some of the first Rust I ever wrote, and I'm kind of embarrassed by it at this point 😅

trustfall_core/src/graphql_query/error.rs
```diff
@@ -88,6 +88,13 @@ pub enum ParseError {
     #[error("Edge {1} specifies a duplicated parameter {0}")]
     DuplicatedEdgeParameter(String, String, Pos),
 
+    #[error(
+        "Found GraphQL query variable definition(s). \
+    These are not necessary since variables are defined implicitly, \
+    and must be removed."
+    )]
+    VariableDefinitionInRootQuery(Pos),
```
obi1kenobi, line 96: I'm confused by the "root query" bit, since there's only one query — we error if the user tries to send more than one query in one document.

I'm curious to hear more about what you had in mind with the name. Also happy to do a quick rename as well:
```suggestion
    VariableDefinitionInQuery(Pos),
```

obi1kenobi (commented):
Sorry you're dealing with my inept use of Rust from back in the day 😅

The code looks good and I'll be very happy to merge it. I just noticed a few spots where it could be more idiomatic and maintainable, and if you don't mind polishing them up while we're touching this part of the repo, I'd really appreciate it. If you'd prefer not to, then no worries — I can merge as-is and tweak them on my own later.

Thanks for putting this together! 🎉

trustfall_core/src/graphql_query/query.rs
devanbenz: Since we are returning a result couldn't I replace `.unwrap()` with a try operator? i.e. `?`

devanbenz: Ope disregard - most of these return an `Option` 😆

trustfall_core/src/graphql_query/error.rs
devanbenz: Ah you're correct - I was sort of following the naming scheme for other variables within the area of code coverage haha. I'll rename.

trustfall_core/src/graphql_query/query.rs
devanbenz: @obi1kenobi these are actually of type `Vec` so I think `.is_empty()` should be fine here. 

Screenshot 2024-04-20 at 9 33 52 AM

obi1kenobi: Ah, sorry, I should have been clearer. We're doing an `.is_empty()` followed by `.first().unwrap()`.

This is the more elegant, panic-free way to do the same:
```rust
if let Some(first_variable_definition) = root_node.variable_definitions.first() {
    return Err(ParseError::VariableDefinitionInRootQuery(first_variable_definition.pos));
}
```

I just don't think I knew about `if let` at the time I wrote this code.

devanbenz:
Need to just fix the tests or see why their failing 😆

devanbenz:
Okay cool - test fixed. Should be good now.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,59 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
...
+                    mult.values()
+                        .next()
+                        .expect("Could not iterate to second value in document.")
+                        .pos,
+                ))
+            } else if let Some(node) = mult.values().next() {
+                parse_operation_definition(node)
+            } else {
+                Err(ParseError::NodeNotFound)
+            }
```
obi1kenobi, line 140: Can we actually trigger this? If not, should it be a panic branch?

I'd love a test case that tries to hit it if you can come up with one. It can be in the next PR.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,59 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
 
     match &document.operations {
         DocumentOperations::Multiple(mult) => {
-            return Err(ParseError::MultipleOperationsInDocument(mult.values().next().unwrap().pos))
+            return if mult.values().len() > 1 {
+                mult.values().next().expect("Could not iterate to first value in document.");
```
obi1kenobi, line 129: I'm confused by this line, since it seems to be a no-op. Did you mean to assign its result and use it in the `Err()` below?

obi1kenobi: Also, the `return if` is not very idiomatic, and I found it surprising -- I even missed it the first time because I didn't expect it there. Could we avoid it by putting the `return` inside each block of the `if`?

devanbenz: I guess the methodology behind this is that its meant to mutate the iterator. The first call to `.next()` will consume the first value (i.e. the first document) the next call to `.next()` would be the second value (the value of the operation that would trigger the `MultipleOperationsInDocument` error.)

trustfall_core/src/graphql_query/query.rs
devanbenz: Yeah I ended up putting in new error type for it but I can definitely make a `panic` then write a test case for it. That shouldn't be a problem what so ever!

trustfall_core/src/graphql_query/query.rs
devanbenz: This may not be a good way to do this though and am happy to refactor :)

devanbenz: @obi1kenobi weird the page didn't load your comment about the `return if` at first 😆 Will take a look - this was actually a suggestion from `clippy` as a linting thing. I wonder if it may be worth it to add a linting rule/exception from clippy to avoid recommending `return if` then?

obi1kenobi: Ah I completely missed that, clearly.

Usually, discarding a return value is done by `let _ = call_whose_value_we_discard()`, with a comment above it explaining what's going on. This will silence any clippy `#[must_use]` errors. In this case, I think the `.expect()` prevented that clippy lint, but I don't think it should have -- so better safe than sorry + follow convention.

obi1kenobi: `nth()` is fine too!

obi1kenobi:
I'm going to be AFK for several hours, so please use the "request review" button once you've had a chance to go through everything and it's time for me to take another look. Thanks for taking care of all this! I appreciate it!

trustfall_core/src/graphql_query/query.rs
obi1kenobi: `return if` is tricky. If the blocks are tiny, it can be very good! In this case, clippy was probably stuck in a local maximum -- but perhaps there's a better way to do it overall.

devanbenz: > `return if` is tricky. If the blocks are tiny, it can be very good! In this case, clippy was probably stuck in a local maximum -- but perhaps there's a better way to do it overall.

This is really good to know.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,57 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
...
+                    mult.values()
+                        .nth(2)
+                        .expect("Could not iterate to second value in document.")
+                        .pos,
+                ))
+            } else if let Some(node) = mult.values().next() {
+                parse_operation_definition(node)
+            } else {
+                // This should be unreachable if someone is using the library correctly
+                panic!("These aren't the droids you're looking for.")
```
devanbenz, line 139: I actually don't think its possible to hit this error with valid GraphQL LOL

obi1kenobi: This is a fun message to write, but it'll be very frustrating if anyone actually hits it. I hate to be the fun police, but could we try to apply some empathy for the poor user that hits this case?
```suggestion
                unreachable!("Found a `DocumentOperations::Multiple()` with no query components. \
                              This shouldn't be possible, and is a bug. Please report it at \
                              https://github.com/obi1kenobi/trustfall/")
```

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,57 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
...
-                    Err(ParseError::UnsupportedQueryRoot("an inline fragment".to_string(), inl.pos))
-                }
-            }
+    let root_selection_set = &root_node.selection_set.node;
+    let root_items = &root_selection_set.items;
+    if root_items.len() != 1 {
+        return Err(ParseError::MultipleQueryRoots(root_items[1].pos));
+    }
+
+    let root_node = root_items.first().unwrap();
```
obi1kenobi, line 171: This is also an opportunity for an `if let Some(root_node) = root_items.first()` and `.nth(2)` block like you put together above.

obi1kenobi (commented):
Nice! Two minor nitpicks; in the interest of merging this PR instead of iterating on it more, let's do this: I can add a suggestion for one of them, apply it, merge the PR, and then if you'd like you can tackle the other one in a subsequent PR.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -125,49 +125,59 @@ fn try_get_query_root(document: &ExecutableDocument) -> Result<&Positioned<Field
...
+                        .expect("Could not iterate to second value in document.")
+                        .pos,
+                ))
+            } else if let Some(node) = mult.values().next() {
+                parse_operation_definition(node)
+            } else {
+                // This should be unreachable if someone is using the library correctly
+                unreachable!("Found a `DocumentOperations::Multiple()` with no query components. \
+                              This shouldn't be possible, and is a bug. Please report it at \
+                              https://github.com/obi1kenobi/trustfall/")
```
obi1kenobi, line 141:
```suggestion
                unreachable!(
                    "Found a `DocumentOperations::Multiple()` with no query components. \
                    This shouldn't be possible, and is a bug. Please report it at \
                    https://github.com/obi1kenobi/trustfall/"
                )
```

PR #601, opened 2024-04-29
feat: small cleanups for query.rs

devanbenz:
- Adds new error type instead of `OtherError` for Unsupported fold directive
- Removes an unwrap and replaces with graceful error handling

### DIFF
--- a/trustfall_core/src/graphql_query/error.rs
+++ b/trustfall_core/src/graphql_query/error.rs
@@ -63,6 +63,9 @@ pub enum ParseError {
     #[error("Unrecognized transform operator: {0}")]
     UnsupportedTransformOperator(String, Pos),
 
+    #[error("Unrecognized fold directive: {0}")]
+    UnsupportedFoldDirective(String, Pos),
+
     #[error("Output name \"{0}\" contains invalid characters: {1:?}")]
     InvalidOutputName(String, Vec<char>, Pos),
 
--- a/trustfall_core/src/graphql_query/error.rs
+++ b/trustfall_core/src/graphql_query/error.rs
@@ -63,9 +63,6 @@ pub enum ParseError {
     #[error("Unrecognized transform operator: {0}")]
     UnsupportedTransformOperator(String, Pos),
 
-    #[error("Unrecognized fold directive: {0}")]
-    UnsupportedFoldDirective(String, Pos),
-
     #[error("Output name \"{0}\" contains invalid characters: {1:?}")]
     InvalidOutputName(String, Vec<char>, Pos),
 
--- a/trustfall_core/src/graphql_query/query.rs
+++ b/trustfall_core/src/graphql_query/query.rs
@@ -172,15 +172,23 @@ fn parse_operation_definition(
         return Err(ParseError::MultipleQueryRoots(root_items[1].pos));
     }
 
-    let root_node = root_items.first().unwrap();
-    match &root_node.node {
-        Selection::Field(positioned_field) => Ok(positioned_field),
-        Selection::FragmentSpread(fs) => {
-            Err(ParseError::UnsupportedQueryRoot("a fragment spread".to_string(), fs.pos))
+    if let Some(root_node) = root_items.first() {
+        match &root_node.node {
+            Selection::Field(positioned_field) => Ok(positioned_field),
+            Selection::FragmentSpread(fs) => {
+                Err(ParseError::UnsupportedQueryRoot("a fragment spread".to_string(), fs.pos))
+            }
+            Selection::InlineFragment(inl) => {
+                Err(ParseError::UnsupportedQueryRoot("an inline fragment".to_string(), inl.pos))
+            }
         }
-        Selection::InlineFragment(inl) => {
-            Err(ParseError::UnsupportedQueryRoot("an inline fragment".to_string(), inl.pos))
-        }
+    } else {
+        unreachable!(
+            "Found a root_node with no items. \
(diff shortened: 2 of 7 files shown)


trustfall_core/src/graphql_query/error.rs
```diff
@@ -63,6 +63,9 @@ pub enum ParseError {
     #[error("Unrecognized transform operator: {0}")]
     UnsupportedTransformOperator(String, Pos),
 
+    #[error("Unrecognized fold directive: {0}")]
+    UnsupportedFoldDirective(String, Pos),
```
obi1kenobi, line 67: Are you able to trigger this error case? (Or is it maybe already triggered by the existing test suite?)

If so, do you feel like this error message is sufficient to explain the problem and guide the user toward resolving it?

obi1kenobi (commented):
Thanks, this looks quite nice!

We probably want to make sure we have test coverage on the new error case. Do you feel like you have a good sense of when that error is triggered, and could explain to a user that sees it how they should tweak their query so it doesn't error anymore?

trustfall_core/src/graphql_query/error.rs
devanbenz: Good callout 👍 I'll go ahead and take a look.

trustfall_core/src/graphql_query/query.rs
```diff
@@ -400,9 +408,11 @@ fn make_field_connection(field: &Positioned<Field>) -> Result<FieldConnection, P
             }
             Some(ParsedDirective::Fold(fold, _)) => break Some(fold),
             Some(ParsedDirective::Transform(_, pos)) => {
-                return Err(ParseError::OtherError(
-                    // TODO: do better
-                    "@transform applied to non-folded edge field".to_string(),
+                return Err(ParseError::UnsupportedDirectivePosition(
+                    "@transform".to_owned(),
+                    "Cannot transform property prior to @fold directive.\
```
obi1kenobi, line 413:
```suggestion
                    "Cannot transform an edge prior to a @fold directive. \
```

trustfall_core/test_data/tests/parse_errors/transform_first_directive.parse-error.ron
```diff
@@ -0,0 +1,4 @@
+Err(UnsupportedDirectivePosition("@transform", "this directive cannot appear as a first directive", Pos(
```
obi1kenobi, line 1:
```suggestion
Err(UnsupportedDirectivePosition("@transform", "Cannot transform an edge prior to a @fold directive. Consider adding @fold before the @transform here.", Pos(
```

trustfall_core/src/graphql_query/query.rs
```diff
@@ -400,9 +408,11 @@ fn make_field_connection(field: &Positioned<Field>) -> Result<FieldConnection, P
...
+                    Please make sure you are using @fold before @transform."
```
obi1kenobi, line 414:
```suggestion
                    Consider adding @fold before the @transform here."
```

trustfall_core/test_data/tests/parse_errors/transform_first_directive.graphql.ron
```diff
@@ -0,0 +1,12 @@
+TestGraphQLQuery (
+    schema_name: "filesystem",
```
obi1kenobi, line 2: There's no way you could have known this (the lack of docs is on me), but the `filesystem` schema is very old and outdated in terms of practices and naming conventions. I've been trying to move all test cases to the newer `numbers` schema since it's newer and better in every way.

You don't have to change this test case if you don't feel like it, we can merge it as-is since I don't want to make you do extra work because of my failure to write docs. Just wanted to mention it so you know for the future.

trustfall_core/src/graphql_query/query.rs
obi1kenobi: Tweaked to avoid implying that `@transform` must _always_ come after `@fold`. That is _currently_ the case because we have exactly one transformation operator, but in principle in the future it doesn't have to be the case.

For example, one could imagine `@transform` being used on properties directly, like `number @transform(op: "flip_sign")` or whatever.

trustfall_core/test_data/tests/parse_errors/transform_first_directive.graphql.ron
```diff
@@ -0,0 +1,12 @@
...
+    query: r#"
+{
+    OriginDirectory {
+        out_Directory_Count @transform(op: "count") @output {
```
obi1kenobi, line 6:
```suggestion
        out_Directory_ContainsFile @transform(op: "count") @output {
```
`out_Directory_Count` is not an edge defined in the schema, so the test might be flaky -- it might report an invalid edge just as well as a failure to use `@transform` correctly.

obi1kenobi (commented):
Looks good! A couple of nitpicks and suggestions to make it even better.

trustfall_core/test_data/tests/parse_errors/transform_first_directive.parse-error.ron
```diff
@@ -0,0 +1,4 @@
+Err(UnsupportedDirectivePosition("@transform", "Cannot transform an edge prior to a @fold directive. Consider adding @fold before the @transform here.", Pos(
+    line: 4,
+    column: 29,
```
obi1kenobi, line 3:
```suggestion
    column: 36,
```

PR #651, opened 2024-08-06
Fix Python build in CI

miikka:
Looks like #650 didn't work.

The problem with the [Linux build](https://github.com/obi1kenobi/trustfall/actions/runs/10269436367/job/28415126602) and [Windows build](https://github.com/obi1kenobi/trustfall/actions/runs/10269436367/job/28415125301) are the breaking changes in #575. From action/upload-artifact's [README](https://github.com/actions/upload-artifact):

>Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

This PR attempts to fix the problem by uploading multiple artifacts and then downloading them all.

The [macOS build](https://github.com/obi1kenobi/trustfall/actions/runs/10269436367/job/28415125898) has two problems:

- https://github.com/actions/setup-python/issues/906#issuecomment-2210869905: Python <3.11 is not supported on the arm64 macOS runners. It seems to work on the older x64 `macos-13` runner?
- https://github.com/PyO3/maturin/issues/2154

Test run here: https://github.com/miikka/trustfall/actions/runs/10270391220 (the final step obviously fails due to lack of permissions)

### DIFF
--- a/.github/workflows/publish-python.yml
+++ b/.github/workflows/publish-python.yml
@@ -52,7 +52,9 @@ jobs:
     steps:
       - uses: actions/download-artifact@v4
         with:
-          name: wheels
+          pattern: wheels-*
+          path: dist
+          merge-multiple: true
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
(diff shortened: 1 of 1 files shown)


obi1kenobi:
Thanks for digging into the issues and making the PR, I appreciate it! 🙏

Out of curiosity, what versions of Python are you using this with? Can I just drop support for <3.11 to simplify the macOS publish flow in the future, or is it still too early for that?

miikka:
Hey, it worked this time! https://pypi.org/project/trustfall/0.1.7/#files

I'm on Python 3.11/3.12 myself. I don't have strong opinions on this, but it does not seem worth it to spend much effort on building wheels for the older versions.

PR #652, opened 2024-08-06
Fix Python CI build for musllinux

miikka:
Follow-up to #651. I'm not an expert with musl, but let's fix the build anyway. There are two commits so you can see what was the fix after uncommenting everything. [Seems to work](https://github.com/miikka/trustfall/actions/runs/10271478510) at least well enough to `import trustfall`.


obi1kenobi (approved):
Amazing, thank you!

PR #654, opened 2024-08-08
Add type hints for native code in pytrustfall

miikka:
When using e.g. `trustfall.Schema` in Python code, mypy and VS Code's Python mode complain about it being untyped. According to [PyO3 docs](https://pyo3.rs/v0.22.2/python-typing-hints.html), the way to add type hints for the native code is to manually maintain `*.pyi` file. Since Trustfall's API is fairly small, this seems feasible.

This PR introduces such a file with enough annotations that a mypy check passes for the library code and the test suite.

The PR also makes the CI run `mypy` to ensure that the type definitions stay up-to-date. I've enabled mypy's `strict` mode since in my opinion it's by default too lenient, but that means adding a few extra annotations.


obi1kenobi:
Thanks! I'll review this when I get back to my laptop. I'm very much in
favor of more & better typing, so thank you for putting this together!

pytrustfall/trustfall/trustfall.pyi
```diff
@@ -0,0 +1,18 @@
...
+
+class InvalidSchemaError(Exception): ...
+class ParseError(Exception): ...
+class ValidationError(Exception): ...
+class FrontendError(Exception): ...
+class InvalidIRQueryError(Exception): ...
+class QueryArgumentsError(Exception): ...
+
+class Schema:
+    def __init__(self, schema_txt: str): ...
```
obi1kenobi, line 14: Possibly a missing annotation here:
```suggestion
    def __init__(self, schema_txt: str) -> None: ...
```

obi1kenobi (approved):
Looks great, thank you so much for putting this together!

PR #706, opened 2024-11-13
remove outdated `#[allow]` in `trustfall_wasm`

suaviloquence:
Was looking at `trustfall_wasm` to see how you made the playground, and I saw this `#[allow]` statement that doesn't seem necessary any more.


suaviloquence:
For reference, it was added in bcfdf23b51b6978b6a03e816b1dd0d30cdc12562

obi1kenobi:
Nice, thank you!

PR #734, opened 2025-01-29
Symlink license file to each crate directory

michel-slm:
This allows them to be picked up by `cargo package` so the license text is included in the distributed crate.

```
trustfall on  add-license via 🦀 v1.80.1
⬢ [fedora-packaging] ❯ for crate in trustfall*; cd $crate; echo $crate; cargo package --list | grep LICENSE.txt; cd -; end
trustfall
    Updating crates.io index
LICENSE.txt
trustfall_core
LICENSE.txt
trustfall_derive
LICENSE.txt
trustfall_filetests_macros
LICENSE.txt
trustfall_stubgen
warning: readme `../README.md` appears to be a path outside of the package, but there is already a file named `README.md` in the root of the package. The archived crate will contain the copy in the root of the package. Update the readme to point to the path relative to the root of the package to remove this warning.
LICENSE.txt
trustfall_testbin
warning: manifest has no description, license, license-file, documentation, homepage or repository.
See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
LICENSE.txt
trustfall_wasm
warning: manifest has no description, license, license-file, documentation, homepage or repository.
See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
LICENSE.txt
```

### DIFF
--- a/trustfall/LICENSE.txt
+++ b/trustfall/LICENSE.txt
@@ -1,0 +1,1 @@
+../LICENSE.txt
--- a/trustfall_core/LICENSE.txt
+++ b/trustfall_core/LICENSE.txt
@@ -1,0 +1,1 @@
+../LICENSE.txt
--- a/trustfall_derive/LICENSE.txt
+++ b/trustfall_derive/LICENSE.txt
@@ -1,0 +1,1 @@
+../LICENSE.txt
--- a/trustfall_filetests_macros/LICENSE.txt
+++ b/trustfall_filetests_macros/LICENSE.txt
@@ -1,0 +1,1 @@
+../LICENSE.txt
--- a/trustfall_stubgen/LICENSE.txt
+++ b/trustfall_stubgen/LICENSE.txt
@@ -1,0 +1,1 @@
+../LICENSE.txt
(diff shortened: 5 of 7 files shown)


obi1kenobi:
Nice, thank you for the fix! If you don't mind me asking, how did you notice this issue? I'm quite curious :)

decathorpe:
(answering for Michel here: we're working on providing a package for cargo-semver-checks for Fedora Linux, and our tooling errors out when crates are detected to not contain license texts they should contain :) )

obi1kenobi:
Thanks for packaging it! I'm hoping no source code or dependency changes were necessary to make the package? 🤞 I'm hoping we can avoid difficult-to-triage issues originating from unexpected release combinations.

decathorpe:
> Thanks for packaging it! I'm hoping no source code or dependency changes were necessary to make the package?🤞

We're still working on it, but so far, I don't think we needed to make any changes.

> I'm hoping we can avoid difficult-to-triage issues originating from unexpected release combinations.

Understandable - this is something we're trying to be very careful about - making only the smallest possible changes, and only when really necessary.

PR #756, opened 2025-03-29
Fix internal test failures caused by incorrect handling of windows paths

CLIDragon:
**Purpose of change**
Test crate generation fails due to inconsistencies between the unix-style path handling code and the automatic windows paths generate by Rusts's `Path` type.

**Describe the solution**
Replace all hardcoded `/` with either `std::path::MAIN_SEPARATOR_STR` or appropriate path building functions. 

I'm not especially attached to the path building method. I think `join()` is nice because it allows path creation in a single line, but there's at least a dozen different ways to do it and if another way is preferred, I'm happy to switch.

**Describe alternatives you've considered**
I think the [typed_path](https://docs.rs/typed-path/latest/typed_path/) crate would solve this, but I think a solution that doesn't involve adding another dependency is better, even if it makes the code a little uglier. 

**Testing**
Tests all pass after the change, where they failed before.

### DIFF
--- a/trustfall_stubgen/src/tests.rs
+++ b/trustfall_stubgen/src/tests.rs
@@ -31,6 +31,9 @@ fn assert_generated_code_compiles(path: &Path) {
 
     let trustfall_lib_str = trustfall_lib_path.display();
 
+    // On Windows, paths are written with unescaped backslashes. Using single
+    // quotes forces literal interpretation of the string, preventing errors
+    // due to invalid escape sequences.
     let cargo_toml = format!(
         "
 [package]
@@ -41,7 +44,7 @@ edition = \"2021\"
 rust-version = \"1.70\"
 
 [dependencies]
-trustfall = {{ path = \"{trustfall_lib_str}\" }}
+trustfall = {{ path = '{trustfall_lib_str}' }}
 
 [workspace]
 "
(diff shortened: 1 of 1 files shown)


trustfall_stubgen/src/tests.rs
```diff
@@ -41,7 +41,7 @@ edition = \"2021\"
 rust-version = \"1.70\"
 
 [dependencies]
-trustfall = {{ path = \"{trustfall_lib_str}\" }}
+trustfall = {{ path = '{trustfall_lib_str}' }}
```
obi1kenobi, line 44: I'm assuming this change was intentional — could you leave a comment above `let cargo_toml` that explains why of all strings in the entire template file, this string in particular is single-quoted?

obi1kenobi (commented):
Looks good! Let's make sure to explain the unusual choice of string in the TOML file though.

Also: since CI isn't running on Windows (and I'm not saying it should be), what do you think our strategy should be with respect to this test and Windows test runs? Obviously "fixed" is better than "broken" so we should merge this PR, but how should we think of Windows going forward and communicate that to users and contributors?

trustfall_stubgen/src/tests.rs
CLIDragon: Sure. On Windows, rust writes path strings in the windows format, i.e. `C:\User\...`, without escaping the backslashes. Since double quotes aren't literal strings in TOML, this causes invalid escape sequence errors.

CLIDragon:
> Also: since CI isn't running on Windows (and I'm not saying it should be), what do you think our strategy should be with respect to this test and Windows test runs? Obviously "fixed" is better than "broken" so we should merge this PR, but how should we think of Windows going forward and communicate that to users and contributors?

Personally, I have no issues with Windows support being a low priority. It's not an uncommon stance in the software community, and supporting Windows can definitely be a maintenance burden. With that said, Windows support is currently fully functional, Rust has quite good cross-platform support, and I can't see any near-future functionality that would require dropping support. 

As far as communication, I think a simple best-effort policy will be enough. Something like:
> Windows support is best-effort only and is not guaranteed. If you find a bug relating to Windows, please report it, but be aware that it might be a while before a fix lands, if it ever does.

obi1kenobi (approved):

obi1kenobi:
Seems reasonable, thanks!

PR #785, opened 2025-07-06
Fix patio11 query to use submittedUrl instead of url

edef1c:
`url` always contains the news.ycombinator.com link, which will always fail the regex match.


obi1kenobi (approved):

edef1c:
Lint failures seem to be on Rust code, which I haven't touched. Presumably an artifact of linter versions not being pinned.

PR #793, opened 2025-07-27
Add repeatable on output, tag and transform.

xd009642:
This PR runs on the assumption that the ability of trustfall-stubgen to generate code for schemas with those present means it can be added to directives. Why am I doing this:

> hmm now that I look at it, output and tag and transform should probably be repeatable too 🤔
> Predrag Gruevski

There's still some schema tests failing that I'll look at tomorrow. I'm hoping that they add something a bit more rigorous so we can be sure this won't break anything but I figured I'd just put the draft PR up in the meantime.


obi1kenobi (approved):
Thank you!

Issue #825, opened 2025-09-25
Use `uv` as package manager for python bindings


deckstose:
After the docs where migrated to use `uv` maybe it's a good idea to align `pytrustfall` as well. Makes for a consistent experience.

Two other things come to my mind:

* `uv` can rebuild on [external file changes](https://docs.astral.sh/uv/reference/settings/#cache-keys), nice for `Cargo.toml` and Rust files.
* `maturin` integrates with `uv` and can use it instead of `pip` for dependency management.

Otherwise, this is not a big step in any direction, rather nice to have.

I'm happy to open a PR.

obi1kenobi:
Sounds good, happy to take a PR!

Probably best to make one change at a time? Say, first switch to `uv` everywhere, and then add config settings to rebuild on external file changes? Let me know what you think.

I was already using `maturin` to develop the Python bindings, but if there's a better way to do it, I'm happy to check it out.

Btw, if you'd like, you're welcome to [join the project Discord](https://discord.gg/tafKHScj)!

PR #828, opened 2025-09-27
Switch python package manager to uv

deckstose:
Bug: https://github.com/obi1kenobi/trustfall/issues/825


pytrustfall/pyproject.toml
```diff
@@ -1,11 +1,13 @@
+[build-system]
+requires = ["maturin>=1.6,<2"]
+build-backend = "maturin"
```
obi1kenobi, line 3: Would you mind keeping this block at the bottom, in its original spot? It's mostly only useful to tools, and I'd like to keep the file order organized by how useful the content is to humans :)

obi1kenobi (approved):
Nice, thank you!

PR #914, opened 2026-04-08
docs: migrate from mkdocs-material to zensical

deckstose:
The creator(s) of mkdocs-material have [created](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/) a successor named [zensical](https://zensical.org). They have also [stopped](https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/) the development of mkdocs-material due to mkdocs, their base, basically stalling in development.

This PR switches to zensical. It would be a drop in replacement but I decided to rewrite the config in the new format.

This also enables a CI pipeline that publishes the docs at https://obi1kenobi.github.io/trustfall/.

I tried to structure the site in the same way as the old site, without editing the content itself.


docs/pyproject.toml
```diff
@@ -1,20 +1,8 @@
 [project]
 name = "trustfall_docs"
 version = "0.0.0"
-description = "Documentation for the Trustfall project"
-authors = [{ name = "Predrag Gruevski" }]
 requires-python = ">=3.11"
```
obi1kenobi, line 4: Should we bump this to 3.14? I don't think we particularly care to support older versions for the docs portion.

obi1kenobi (approved):
Thank you so much!

docs/pyproject.toml
deckstose: Yes although uv might pick the latest stable version already in its action

deckstose:
Alao, you need to enable pages for the repository.

PR #928, opened 2026-05-15
Update weather example for May 2026 CSV format

### LINKED ISSUE #927: Weather example panics when run
It appears that the example expects a specific METAR CSV format. This has changed since the code was first written. Panics include:
```
thread 'main' (15181) panicked at trustfall/examples/weather/main.rs:70:22:
internal error: entered unreachable code
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:80:14
   2: core::panicking::panic
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:150:5
   3: weather::read_metar_data
             at ./examples/weather/main.rs:70:22
   4: weather::run_query
             at ./examples/weather/main.rs:89:16
   5: weather::main
             at ./examples/weather/main.rs:172:17
   6: core::ops::function::FnOnce::call_once
             at /home/james/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

jameshu15869:
Previously, the weather example made the following incorrect assumptions about the CSV from aviationweather.gov:

1. The file would contain a CSV prefix.
2. elevation_m would always be a valid f64

This PR removes handling for the CSV prefix and deserializes `null` to None.

Fixes #927

### DIFF
--- a/trustfall/examples/weather/main.rs
+++ b/trustfall/examples/weather/main.rs
@@ -56,24 +56,15 @@ fn read_metar_data() -> Vec<MetarReport> {
 
     let mut buf = String::new();
 
-    // strip the CSV prefix and the header row
-    let prefix_len = 6;
-    for _ in 0..prefix_len {
-        reader.read_line(&mut buf).unwrap();
-
-        match buf.as_str().trim() {
-            "No errors" | "No warnings" | "data source=metars" | METAR_DOC_HEADER_ROW => {}
-            data => match data.split_once(' ') {
-                Some((left, right)) if right == "ms" || right == "results" => {
-                    assert!(left.chars().all(|x| x.is_ascii_digit()));
-                }
-                _ => unreachable!(),
-            },
-        }
-
-        buf.truncate(0);
+    // Strip header row
+    reader.read_line(&mut buf).unwrap();
+    if buf.as_str().trim() != METAR_DOC_HEADER_ROW {
+        unreachable!("Expected CSV to begin with only the header ROW")
     }
 
+    // We cannot use has_headers(true) because the CSV has duplicate column headers
+    // for sky_cover and cloud_base_ft_agl, which we manually handle.
+    // (See [`crate::metar::CsvMetarReport`])
     let mut csv_reader = csv::ReaderBuilder::new().has_headers(false).from_reader(reader);
 
     let metars: Vec<MetarReport> =
--- a/trustfall/examples/weather/metar.rs
+++ b/trustfall/examples/weather/metar.rs
@@ -79,6 +79,7 @@ pub(crate) struct CsvMetarReport {
     pub(crate) vert_vis_ft: Option<i32>,
 
     pub(crate) metar_type: Option<String>,
+
     // As of May 2026, METAR may contain `null` in this field
     #[serde(deserialize_with = "csv::invalid_option")]
     pub(crate) elevation_m: Option<f64>,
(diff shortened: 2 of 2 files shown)


trustfall/examples/weather/main.rs
```diff
@@ -52,29 +41,9 @@ fn read_metar_data() -> Vec<MetarReport> {
...
-                }
-                _ => unreachable!(),
-            },
-        }
-
-        buf.truncate(0);
-    }
-
-    let mut csv_reader = csv::ReaderBuilder::new().has_headers(false).from_reader(reader);
+    let mut csv_reader = csv::ReaderBuilder::new().has_headers(true).from_reader(reader);
```
obi1kenobi: `.has_headers(true)` appears to break the deserialization of the `sky_cover` and `cloud_base_ft_agl` fields, which now incorrectly return `None` here.

Upstream those fields have the same column name across 4x repeats, but our CSV library doesn't support that. We use our explicit header row to assert that the CSV format is as expected, and then use positional parsing to split them into separate fields (with suffixes `_1, _2, _3, _4`) to allow parsing the file and combining those into list fields manually.

Could you revert to `.has_headers(false)` and retain the original shape of the code while applying a more minimal fix, and perhaps also adding an explanatory comment to explain the above for the future? I agree it isn't obvious 😅

obi1kenobi (commented):
Hi! Thanks for finding the issue and opening a PR to help fix it.

There's a small bug here. If you don't mind fixing it up, I'd be happy to merge this. No pressure, I can also fix it too -- just wanted to give you credit in the commit history if you'd like it :)

trustfall/examples/weather/main.rs
jameshu15869: Thanks for the catch! I reverted those changes. Just curious, how did you catch this? The example seemed to run fine on my local machine - are there parts of the repo I"m missing?

obi1kenobi: This is the full repo, no missing parts.

I found it curious that I had originally felt the need to inline the CSV header row and then use `.has_headers(false)`. I am not the kind of person to have done that without a reason, so I noted the likely [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence) situation, wished I had left a note with my reasoning, and then asked GPT 5.5 in Codex to review the code locally. It correctly pieced together what happened, and then I then verified its conclusion manually and wrote up my review comment.

trustfall/examples/weather/metar.rs
```diff
@@ -79,6 +79,8 @@ pub(crate) struct CsvMetarReport {
     pub(crate) vert_vis_ft: Option<i32>,
 
     pub(crate) metar_type: Option<String>,
+    // As of May 2026, METAR may contain `null` in this field
```
obi1kenobi, line 83:
```suggestion

    // As of May 2026, METAR may contain `null` in this field
```

obi1kenobi (approved):
Looks great! Just adding one newline to visually separate the more complex field from its predecessor, and merging.

Thanks again for catching this and fixing it! If you end up building something with Trustfall, I'd love to hear about it too.

PR #936, opened 2026-06-14
Update ron to 0.12

musicinmybrain:
https://github.com/ron-rs/ron/blob/v0.12.1/CHANGELOG.md

A source-code change was required in `trustfall_testbin` to get from `ron` 0.8 to 0.9. I confirmed that `cargo test --workspace` still passes.


obi1kenobi (approved):
Thank you!

PR #946, opened 2026-06-22
Change docs to use hackernews schema.

xd009642:
This allows the docs to become more interactive by linking to the trustfall playground and running queries for users. 

This is currently a fairly minimal switchover from the fake-twitter to hackernews and the semantic differences between the two APIs. There's a few "run this in the playground" links where relevant and the schemas match.


obi1kenobi (approved):

obi1kenobi:
Thanks! Sorry I'm only just now working through the backlog.

PR #955, opened 2026-07-20
add `repeatable` to all `@transform` definitions

ivelieu:
As discussed on the discord, some of the `@transform` definitions (and corresponding test outputs) were missing `repeatable`. I'm not getting any problems running `cargo test` locally.


obi1kenobi:
Thanks!

PR #974, opened 2026-09-02
Add native async adapters and execution

philocalyst:
## Summary

Welcome to part TWO of "huge and large changes to trustfall", covering async. It is still best read after the base PR: this is a real stream engine, not an async wrapper around the synchronous API.

## Public API

- `AsyncAdapter` and `AsyncBasicAdapter` are the direct path: resolvers use values and `Stream`s, and `execute_query_async()` yields plain rows.
- `FallibleAsyncAdapter` is the explicit error path. `try_execute_query_async()` yields `Result<Row, ExecutionError<E>>` without contaminating direct adapters with `Infallible`, `Ok`, or boxed `map(Ok)` plumbing.
- Both APIs are runtime-agnostic, `!Send`-friendly, and use pinned boxed stream aliases only at the boundary where type erasure is actually useful.
- `AsyncBasicAdapter` keeps the small `&str` / automatic-`__typename` surface. Implement `AsyncAdapter` directly only when resolver metadata matters.

## Semantics and implementation

- Resolver streams preserve context order. A failed row retains its place and unrelated rows continue.
- Errors before an edge stream starts and errors while neighbors stream are represented at the appropriate stream boundary.
- The synchronous adapter is projected onto the same kernel with a ready-only bridge that preserves lazy batch resolution and error positions.
- Ordered buffered helpers handle concurrent per-context work without making a runtime choice for users.
- Filtering, recursion, and fold output materialization were reduced to smaller stream stages. Fold output seeding is now a recursive shape operation instead of an inlined queue walk; recursive piggybacks STAYS lazy.

## Documentation

- The Rust Adapters guide gets the async continuation: direct `AsyncAdapter` / `AsyncBasicAdapter`, explicit `FallibleAsyncAdapter`, and the two edge-error boundaries.

## Migration

Existing synchronous adapters remain on #973. Async adapters that cannot fail implement `AsyncAdapter` or `AsyncBasicAdapter` directly. Async adapters that can fail implement `FallibleAsyncAdapter` and call `try_execute_query_async()`. The live demo now uses `FallibleDemoAdapter` with `try_execute_query()`, so remote Hacker News starting-edge failures are actual result rows rather than a pretend fallible interface. (yay consistency)

## Stack

Depends on #973, which establishes the direct and fallible synchronous contracts. Review that PR first.

## Validation

- Core async and synchronous differential coverage
- Workspace tests, clippy, rustdoc, and formatting
- Generated-stub fixtures and WASM build


obi1kenobi:
I appreciate the enthusiasm, but I'm not sure what the goal is here — especially given you haven't communicated with me at all before opening 10k lines worth of PRs.

I have plenty of AI tokens myself, so merely contributing tokens is not helping the project. Beyond that, I don't see how these PRs are meant to help.

PR #976, opened 2026-09-05
Execute synchronous queries on a stream kernel

philocalyst:
## Summary

WELCOME to the (attempt 2) PR stack for tackling two critical parts of Trustfall's API: errors, and async.

For this to make sense, there needs a beautiful foundation. This PR is said foundation.

Prior, the execution stages used to be an iterator pipeline, built inline to `execution.rs`. Well, now they are a runtime-agnostic stream. This makes the sync API a mere projection on top of it. This means, for now: it's perfectly compatible (this is not a fact that changes really throughout the PRs, but it's very true here). That being said, this PR is a slippery slope, I can't imagine you would want to merge this and not the rest. So please skim the other PRs when you can! Kind of a package deal here.

## What moved where

- `engine.rs` is our kernel: component traversal, edge expansion, and fold materialization! recursion and filtering go into `engine_filter.rs` and `engine_recurse.rs`. (should we make this a folder?)
- `async_adapter.rs` defines the kernel's own `Stream`-native adapter contract. It's private. The public async API is a few PRs away yet!
- `sync_adapter.rs` bridges the classic sync `Adapter` onto the new contract. The bridge doesn't cheat: resolvers still receive whole lazy context batches rather than collected vectors or one-item calls, and results don't lose lazy. A suspended stream here is a frontend bug, no reason to spin or block.
- `execution.rs` keeps `interpret_ir` and the fold-count bounds and is now a thin frontend.
- `error.rs` adds `ExecutionError`, the kernel's error channel. The synchronous `Adapter` cannot fail, so its error type is uninhabited and the `Result` is discharged at the frontend.

`DynamicallyResolvedValue` resolves tagged values directly rather than calling back into the old `execution.rs` internals, which no longer exist.

## Why is there an async trait and an error type in the *foundation* PR

Well, it would be nice to split them out, but it's very highly coupled to the rest of the changes here, and I need my CI checks to pass, heh :) 

## Why streams, given that nothing is async yet

Well, it's kind of the one substrate I think that doesn't cause complexity to actually blow up. I think you could PROBABLY do an async frontend on top of an iterator engine? But then it would fake lazy. This is cleaner I swear

## Migration

None. `Adapter`, `BasicAdapter`, `execute_query()`, and `interpret_ir()` all keep their existing signatures and semantics :)

## Stack

This is the base of the whole stack. Reviewing it first makes the rest much smaller.

1. **YOU HERE** — stream execution kernel
2. #977 — first-class adapter errors
3. #978 — fallible adapter docs and demo
4. #979 — row-ordered error propagation
5. #980 — native async adapters
6. #982 — async ergonomics
7. #981 — async docs and the weather example

And BE AWARE: They look wayyy more scarier than they actually are, showing a cumulative diff. I think the average PR is around a 700 line diff, which may still be a lot (it used to be worse, don't check my closed PRs...) 

## Validation

(CI CHECKS)
- Workspace tests, clippy, rustdoc, and formatting
- Fuzz workspace clippy, rustfmt, and rustdoc
- New synchronous-bridge tests covering batch integrity and result laziness


philocalyst:
@obi1kenobi ready for feedback! actually crazy set of changes, so I'm open to a totally different direction. Tried to make it feel managable, happy to answer any questions.

